PR #4, opened 2021-04-18
Adds transparency to AudioReactiveObject (assuming shader support)

seanidzenga:
I'd like to add transparency support to AudioReactiveObjects - I have some club lights (lasers) in my project that are just long cylinders that I'd like to "turn on" to the beat so I figured transparency would be the best way to implement it and also figured it'd be good to add this to the main project.


seanidzenga:
Sorry I found some bugs lemme sort those out - sorry for being hasty.

seanidzenga:
false alarm, the bug I saw was from something unrelated, good to go

PR #52, opened 2021-05-26
Added a Namespace to everything and add a few helpful things to CGInclude that I felt were missing

Xiexe:
- Added namespace to all scripts to prevent any conflicts
- Added defines to quickly sample from Bass, Low Mids, High Mids, and Trebles
- Added function to test if Audio Link Texture is available for Non Surface shaders to CGInc `AudioLinkIsAvailableNonSurface()`


Xiexe:
Adding the namespace makes it look like i changed more than i did but it eez how it eez /shrug

PR #59, opened 2021-05-28
Add smoothfont test to attic.

cnlohr:


llealloo:
Let's keep this off the main repo until it can live downstairs

PR #72, opened 2021-06-02
Add WIP changelog and update discrepancies in Docs

pema99:
Please read:
- Changelog is a skeleton and needs to be filled out if we go with it.
- I fixed the major discrepancies in the docs, however:
    - A few of the functions in AudioLink.cginc are still not described (such as ALDecodeDataAsFloat)
    - The table 'Table for does it make sense to index with?' maybe should be updated, I'm not sure. I'll let Charles look.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Changelog
+## 0.2.5 - Date here
+### Breaking changes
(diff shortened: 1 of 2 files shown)


pema99:
https://keepachangelog.com/en/1.0.0/

PR #100, opened 2021-06-21
Fixed Player Info pixel refrence in README.md

Shadowriver:
Colors in player info was shifted and removed "reserved" as all empty spaces are reserved anyway


llealloo:
please resubmit to dev

PR #127, opened 2021-09-27
Replace previous UI sprite atlas with manual atlas.  

jaquadro:
This atlas does not require changing project sprite packer settings.
Applied to controller and mini player.
Should more reliably reduce draw calls on controller (ideally down to 2 calls).


pema99:
Checked out your branch. Looks good to me.

PR #146, opened 2021-11-14
Add AudioLink Couch.

cnlohr:


cnlohr:
* Add export script.

 * Move "wheel" model that's unused out of export paths.
 * Make all paths absolute to avoid copy-paste errors by shader makers.
 * Remove LOD to remove confusion among shader makers.
 * Re-export default AudioLink prefab with AudioExport turned off.
 * Bump version number to 0.2.7

pema99:
nice script

PR #149, opened 2021-11-17
Fix bug where mirror causes global AudioTexture to be replaced

pema99:
As discussed on discord.

This fix updates the AudioLink prefab in the following ways:
- The camera is now using Legacy Vertex Lit rendering
- The AudioLinkExport shader has been rewritten. It will now be ignored by cameras on Forward mode (including the main and mirror camera)
- Removed replacement shader code, it is no longer needed and seems to cause interference. As noted, Forward cameras will ignore the AudioLinkExport shader now.

I had to update various assets and scenes to reflect the changes to the prefab. If you wanna look at the diff, I suggest looking at just the first of 2 commits.


Float3:
AudioLinkAvatar.prefab also needs to be updated!

PR #154, opened 2021-12-06
Minimal fix for non-networked usage

yewnyx:
A bit of an experiment with URP

image

### DIFF
--- a/AudioLink/Scripts/AudioLink.cs
+++ b/AudioLink/Scripts/AudioLink.cs
@@ -113,6 +113,8 @@ public class AudioLink : UdonSharpBehaviour
 #if VRC_SDK_VRCSDK2 || VRC_SDK_VRCSDK3
         private double GetElapsedSecondsSince2019() { return (Networking.GetNetworkDateTime() - new DateTime(2020, 1, 1) ).TotalSeconds; }
         //private double GetElapsedSecondsSinceMidnightUTC() { return (Networking.GetNetworkDateTime() - DateTime.UtcNow.Date ).TotalSeconds; }
+#else
+        private double GetElapsedSecondsSince2019() { return 0; }
 #endif
 
         // Fix for AVPro mono game output bug (if running the game with a mono output source like a headset)
@@ -210,6 +212,13 @@ public class AudioLink : UdonSharpBehaviour
                 Networking.IsInstanceOwner?1.0f:0.0f,
                 0 ) );
 
+            #else
+                audioMaterial.SetVector("_PlayerCountAndData", new Vector4(
+                0,
+                0,
+                0,
+                0 ) );
+            #endif
             _FPSCount = 0;
             _FPSTime++;
 
@@ -229,7 +238,11 @@ public class AudioLink : UdonSharpBehaviour
             }
 
             // Finely adjust our network time estimate if needed.
+            #if VRC_SDK_VRCSDK2 || VRC_SDK_VRCSDK3
             int networkTimeMSNow = Networking.GetServerTimeInMilliseconds();
+            #else
+            int networkTimeMSNow = (int)(Time.time*1000.0f);
+            #endif
             int networkTimeDelta = networkTimeMSNow - _networkTimeMS;
             if( networkTimeDelta > 3000 )
             {
@@ -247,7 +260,6 @@ public class AudioLink : UdonSharpBehaviour
                 _networkTimeMS += networkTimeDelta/20;
             }
             //Debug.Log( $"Refinement: ${networkTimeDelta}" );
-            #endif
         }
 
         private void Update()
--- a/AudioLink/Scripts/AudioLink.cs
+++ b/AudioLink/Scripts/AudioLink.cs
@@ -1,6 +1,8 @@
 ﻿using UnityEngine;
 #if VRC_SDK_VRCSDK2 || VRC_SDK_VRCSDK3
 using VRC.SDKBase;
+#else
+using UnityEngine.Rendering;
 #endif
 using UnityEngine.UI;
 using System;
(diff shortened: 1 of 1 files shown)


AudioLink/Scripts/AudioLink.cs
```diff
@@ -1,8 +1,11 @@
 ﻿using UnityEngine;
-using VRC.SDKBase;
 using UnityEngine.UI;
 using System;
 
+#if VRC_SDK_VRCSDK3
```
pema99, line 5: Nice, is this define documented somewhere though?

AudioLink/Shaders/AudioLinkTestTime.shader
```diff
@@ -2,6 +2,7 @@
 {
     Properties
     {
+        _AudioTexture("_AudioTexture", 2D) = "white" {}
```
pema99, line 5: What is this for? Are you passing in the rendertexture explicitly? Not sure how properties will interact with existing named grabpasses

AudioLink/Shaders/HackyExport.shader
```diff
@@ -0,0 +1,112 @@
+﻿Shader "Unlit/HackyExport"
```
pema99, line 1: Where is this shader being used? Seems like something that should go in AudioLinkSandbox?

pema99: Also, what is this actually for? It seems like you are reexporting the named grabpass, but we already have a setup for exporting it. Why not use that?

AudioLink/Shaders/HackyExport.shader
```diff
@@ -0,0 +1,112 @@
...
+                // apply fog
+                UNITY_APPLY_FOG(i.fogCoord, col);
+                col = 1.;
+                return col;
+            }
+            ENDCG
+        }
+        GrabPass
+        {
+            "_AudioTexture"
```
pema99, line 62: If the goal is of this shader is just to export some texture globally, one can use the camera trick we already use for the main AudioLink prefab like I mention above.

AudioLink/RenderTextures/dont_use_audio_link.renderTexture
```diff
@@ -0,0 +1,37 @@
+%YAML 1.1
```
pema99, line 1: Also here, the name implies to me that this is something that should go into AudioLinkSandbox, and not the main folder

AudioLink/RenderTextures/Materials/rt_AudioLink.mat
```diff
@@ -0,0 +1,81 @@
...
+%TAG !u! tag:unity3d.com,2011:
+--- !u!21 &2100000
+Material:
+  serializedVersion: 6
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_Name: rt_AudioLink
+  m_Shader: {fileID: 4800000, guid: 933532a4fcc9baf4fa0491de14d08ed7, type: 3}
```
pema99, line 11: Changes to these .mat files are always a bit opaque. Did you actually change anything meaningful here, if so what?

AudioLink/Scripts/AudioLink.cs
yewnyx: https://docs.vrchat.com/docs/sdk-troubleshooting#im-getting-errors-related-to-sdk3-or-udon-in-a-project-using-sdk2-or-vrc_spatialaudiosource-in-a-project-using-sdk3

pema99: ah, missed that blurb, thanks

cnlohr:
I will start poking at this tomorrow, I think.  I don't expect to convert everything over, and this is just a vague guide... But I have some ideas of how to accomplish this.

AudioLink/Shaders/AudioLinkTestTime.shader
Float3: if _AudioTexture is defined as a Property it will be sampled instead of the named grabpass
this is only viable for non-UGC, by just assigning the texture, or if you set the property via script

pema99:
fyi https://github.com/llealloo/vrc-udon-audio-link/pull/155

AudioLink/Shaders/AudioLinkTestTime.shader
cnlohr: @yewnyx do not commit this change.

AudioLink/Shaders/HackyExport.shader
cnlohr: This shader ended up not working, don't commit it.

AudioLink/Shaders/HackyExport.shader
cnlohr: With URP, this ended up being a dead end.

AudioLink/RenderTextures/dont_use_audio_link.renderTexture
cnlohr: This was also part of the dead-ended path.

AudioLink/RenderTextures/Materials/rt_AudioLink.mat
cnlohr: Please revert this change.

AudioLink/Shaders/HackyExport.shader
pema99: Dead end as in this file should not be included, or dead end as in the existing functionality didn't work?

yewnyx:
Removed the second commit and tested - the shader with the debug info is now zeroed out, but the normal audiolink shader still works.
image

yewnyx:
Looks like the main thing needed to get it working in URP is to add in
```
_AudioTexture("_AudioTexture", 2D) = "white" {}
```
to the `Properties { ... }` part of the shader, then select the audiolink render texture in the audiolinked material.

PR #155, opened 2021-12-09
Fixed Compilation without VRC and Console warnings

Float3:
compilation tested with UdonSharp + SDK, just Avatar SDK, and without either SDK

though this doesn't fix the issue that script references would still be broken without UdonSharp


AudioLink/Scripts/AudioLink.cs
```diff
@@ -100,14 +102,18 @@ public class AudioLink : MonoBehaviour
         private double _elapsedTimeMSW = 0;
         private int    _networkTimeMS;
         private double _networkTimeMSAccumulatedError;
+#pragma warning disable 414 ///The private field is assigned but its value is never used
```
pema99, line 105: Is this caused by the introduced if guards? If so, you should just guard this variable as well.

pema99: If not and it's never used, remove it

AudioLink/Shaders/AudioLink.cginc
```diff
@@ -199,7 +199,7 @@ float4 AudioLinkGetAmplitudeAtFrequency(float hertz)
 float AudioLinkGetAmplitudeAtNote(float octave, float note)
 {
     float quarter = note * 2.0;
-    return AudioLinkLerpMultiline(ALPASS_DFT + float2(octave * AUDIOLINK_EXPBINS + quarter, 0));
+    return AudioLinkLerpMultiline(ALPASS_DFT + float2(octave * AUDIOLINK_EXPBINS + quarter, 0)).r;
```
pema99, line 202: Good catch, but this is bad. The function should just return float4 like the above one

pema99:
Did you use latest U# to test?

Also, have you checked if thus actually compiles without UdonSharp now? I think some class definitions might still be needing guards (UdonSharpBehaviour VS MonoBehaviour)

Float3:
> Did you use latest U# to test?
> 
> Also, have you checked if thus actually compiles without UdonSharp now? I think some class definitions might still be needing guards (UdonSharpBehaviour VS MonoBehaviour)

I used UdonSharp v1.0.0b8
I didn't get any compilation errors in a project with just this (no U#, no Avatar SDK, no World SDK)
Script references are broken though and I didn't test functionality

yewnyx:
> Yewnyx submitted a PR which does something similar to this. Depending on what lands first, you might have to address some merge conflicts, just heads up

I'm not territorial, I don't mind dropping mine

AudioLink/Scripts/AudioLink.cs
Float3: > Is this caused by the introduced if guards? If so, you should just guard this variable as well.

no this was happening before the if guards, it only happens if UDON is not defined so only in avatar projects
 #if guarding it gave me compiler errors
I will check this again

AudioLink/Shaders/AudioLink.cginc
Float3: will address this in the next commit

AudioLink/Scenes/AudioLink_ExampleScene/Scripts/MirrorToggle.cs
```diff
@@ -16,7 +18,9 @@ void Start()
         
     }
 
+#pragma warning disable 108,114 ///hides inherited member
```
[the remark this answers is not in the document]
Float3: I'll try adding these

AudioLink/Scripts/AudioLink.cs
Float3: @pema99 if I #if guard this with UDON so that it doesn't throw the error in avatar projects, then it won't be available for future non-UDON projects

Float3:
> Yewnyx submitted a PR which does something similar to this. Depending on what lands first, you might have to address some merge conflicts, just heads up

I checked it out.



> > Yewnyx submitted a PR which does something similar to this. Depending on what lands first, you might have to address some merge conflicts, just heads up
> 
> I'm not territorial, I don't mind dropping mine

it definitely looks like you are planning to do quite a bit more than me here, I am just fixing some things that were bothering me

yewnyx:
I'll rebase it after and talk to cn, nbd.

AudioLink/Scripts/AudioLink.cs
pema99: Ok, fair enough

pema99:
> > Did you use latest U# to test?
> > Also, have you checked if thus actually compiles without UdonSharp now? I think some class definitions might still be needing guards (UdonSharpBehaviour VS MonoBehaviour)
> 
> I used UdonSharp v1.0.0b8 I didn't get any compilation errors in a project with just this (no U#, no Avatar SDK, no World SDK) Script references are broken though and I didn't test functionality

Oh, U# 1.0. I've been trying to stick to pre-1.0 versions since in my experience recompiling assets with newer versions can break them for older ones. 1.0 isn't _really_ released yet. I'm gonna verify if I have errors when opening the project with latest released version.

Float3:
I initially used the latest release package on the U# GitHub but AL was throwing errors so I switched

pema99:
Ok, I tested with 0.20 with your changes and it looks like everything is in order. Thanks for the contribution.

PR #163, opened 2022-03-03
Add header guard to AudioLink.cginc

PiMaker:
This is a public facing API, and as such should have a header guard, so that modular shaders can include the file wherever necessary without stepping on each other's toes.

As a practical example, adding AudioLink support to LTCGI usually breaks shaders when both are used in combination on one material, since `LTCGI.cginc` also includes `AudioLink.cginc`.


pema99:
Very nice, thanks.

PR #164, opened 2022-03-04
Fixed vertical uv flip of audio texture on Quest

Shadowriver:
Left comment describing issue in the code for future code lurkers


pema99:
Thanks :)

PR #169, opened 2022-04-13
Fixed NullReferenceException in LinkAll.

Nestorboy:
Fixed a NullReferenceException that would occur if you have an empty UdonBehaviour when trying to link all sound reactive objects.


pema99:
Can you resubmit this to the dev branch?

PR #174, opened 2022-05-02
Add UTC (Unix Epoc Indexed) Timestamps

cnlohr:


pema99:
Works great on my side. Let's consider a follow up with some helper methods

PR #175, opened 2022-05-02
Update and add some audiolink note functions

Float3:


pema99:
Use float4 throughout, people may want different parts of the DFT data

PR #176, opened 2022-05-06
Change fade defaults to be more responsive

DomNomNom:
At a lot of events I've seen the host or DJ change these sliders to be similar to this such that the visuals are more responsive to the music.

Even though users can change these, I think it's worth spending a little time to have good defaults.

Before:
image

After: 
image


pema99:
I'm fine with this.

PR #178, opened 2022-05-07
Cleanup repeated uint encoding code.

DomNomNom:
Could've maybe converted the specialized implementations to use this but it seems like someone had fun with optimization :)


pema99:
thanks as always

PR #179, opened 2022-05-07
fix pixelated logo on controller

DomNomNom:
Filter mode used to be "Point", now it's "Bilinear".

context: llealloo's suggestion https://discord.com/channels/840894950011043841/840899396216291329/972195518174330890


pema99:
Very nice

PR #180, opened 2022-05-08
Adjust null checks for the audioSource

techanon:
This prevents AudioLink script from crashing when an audioSource is not present, as it might only be assigned during runtime instead of in editor.  
Add a warning to notify users about the missing audioSource in case it was unintentional.


llealloo:
Seems like a good qol improvement at no cost, thanks!

Not sure what the change is in Update?

techanon:
The change in Update is moving another reference to audioSource inside an already existing null check
image

pema99:
This is totally reasonable. Thanks

PR #181, opened 2022-05-11
Add ThemeColor controller

DomNomNom:
Re-purposes the bottom of the controller to show current theme colors 

Adds a selection for themecolor mode and displays current theme color:
image

Adds a controller to change the custom theme colors
image


pema99:
This PR shouldn't be against master, please retarget to dev. More pressingly, it sort of looks like you based your branch of an old version of master? The diff is huge has conflicts with dev.

DomNomNom:
Fixed the base / conflicts

AudioLink/Scripts/AudioLinkController.cs
```diff
@@ -64,7 +67,21 @@ void Update()
 
             void Start()
             {
-                if (audioLink == null) Debug.Log("Controller not connected to AudioLink");
+                if (audioLink == null) {
```
pema99, line 70: I won't block the PR because of it, but again, nice to keep the bracing style consistent.

AudioLink/Scripts/AudioLinkController.cs
```diff
@@ -64,7 +67,21 @@ void Update()
...
+                    Debug.LogError("Controller not connected to AudioLink");
+                    return;
+                }
+                if (themeColorController == null) {
+                    // This is here in case someone upgraded AudioLink, which updates
+                    // everything in the prefab, but not the outermost properties of the prefab.
+                    // TODO: Double check that this is how upgrading ends up working.
```
pema99, line 77: Did you verify this yourself yet, or shall I do it?

AudioLink/Scripts/AudioLinkController.cs
```diff
@@ -64,7 +67,21 @@ void Update()
...
+                    Debug.Log("AudioLinkController using fallback method for finding themeColorController");
+                    Transform controllerTransform = transform.Find("ThemeColorController");
+                    themeColorController = (UdonBehaviour) controllerTransform.GetComponent(typeof(UdonBehaviour));
```
pema99, line 80: This looks like a nullref exception waiting to happen in case "ThemeColorController" couldn't be found, or if it doesn't have an UdonBehaviour component. Perhaps add some null checks?

AudioLink/Scripts/AudioLinkController.cs
```diff
@@ -124,14 +154,6 @@ public void UpdateSettings()
...
                 audioLink.SetProgramVariable("threshold2", threshold2Slider.value);
                 audioLink.SetProgramVariable("threshold3", threshold3Slider.value);
-                audioSpectrumDisplay.SetFloat("_X0", x0Slider.value);
-                audioSpectrumDisplay.SetFloat("_X1", x1Slider.value);
-                audioSpectrumDisplay.SetFloat("_X2", x2Slider.value);
-                audioSpectrumDisplay.SetFloat("_X3", x3Slider.value);
-                audioSpectrumDisplay.SetFloat("_Threshold0", threshold0Slider.value);
-                audioSpectrumDisplay.SetFloat("_Threshold1", threshold1Slider.value);
-                audioSpectrumDisplay.SetFloat("_Threshold2", threshold2Slider.value);
-                audioSpectrumDisplay.SetFloat("_Threshold3", threshold3Slider.value);
```
pema99, line 134: I don't mind at all, but I'm just curious. Why'd you move this?

AudioLink/Scripts/ThemeColorController.cs
```diff
@@ -0,0 +1,147 @@
...
+    #if UDON
+        using UdonSharp;
+        using VRC.Udon;
+
+        [UdonBehaviourSyncMode(BehaviourSyncMode.Manual)]
+        public class ThemeColorController : UdonSharpBehaviour
+        {
+            [UdonSynced] private int themeColorMode;
+
+            [UdonSynced] public Color[] customThemeColors = new Color[4]{
```
pema99, line 18: Wait, you can sync this? I wasn't aware o.o

pema99:
This looks great. Thanks for doing it, the theme colors are feeling like a real feature now. I've added a few minor comments.

About the networking code added to support the customizable theme colors, have you tested it? In my experience it can be quite tricky to get Udon networking code right, so I'd like to know if you'v verified it with multiple VRC clients.

AudioLink/Scripts/AudioLinkController.cs
DomNomNom: I have not yet verified this. I'd appreciate it if you have time

AudioLink/Scripts/AudioLinkController.cs
DomNomNom: Done. Added null-checks.

AudioLink/Scripts/AudioLinkController.cs
DomNomNom: because it does not depend on `audioLink`, which is now gated behind a null-check

AudioLink/Scripts/ThemeColorController.cs
DomNomNom: :D
From [Udon docs](https://docs.vrchat.com/docs/udon-networking#requestserialization):

> You can sync variables which are arrays for the following types:
> bool, char, byte, sbyte, short, ushort, int, uint, long, ulong, float, double, Vector2, Vector3, Vector4, Quaternion, Color and Color32.

AudioLink/Scripts/AudioLinkController.cs
DomNomNom: Fixed. I'm too used to having a linter auto-fix my formatting ^^

pema99:
Seems like the upgrade logic works fine

PR #182, opened 2022-05-14
Video theme color

DomNomNom:
In-progress...


pema99:
Are you interested in doing this, Dom? Or is it dead

DomNomNomVR:
I think what demotivated me from going forward with this was that there didn't seem to be an universal way of getting the current video frame in VRC as there's two base video components that use different methods and many prefabs that wrap those.
it seems easier in CVR where it's always a render texture that could be dragged in

What are you thoughts on this?

Thank you for the poke about this btw.

pema99:
Yeah, indeed, getting the video texture is annoying as hell. I think you'd basically be forced to add support for video players individually.

DomNomNomVR:
In that case, I don't particularly mind if this gets closed to keep things clean. I might develop this on the CVR side at some point in the future.

pema99:
Let me know if you do. I might be tempted to salvage it back into something that works for a few select video players

pema99:
Just reopen it if you want

PR #188, opened 2022-08-13
Creates VPM-Compatible Package

momo-the-monster:
Replaces shader include paths with new Packages paths
Separates scripts into Runtime / Editor and adds Assembly Definition Files
Adds Package Manifest


Float3:
The AudioLinkSandbox folder should probably either be moved into the assets folder or be part of the package.

pema99:
AudioLinkSandbox is for stuff we don't want to distribute, so it shouldn't be in the package. It can go in the project, I think.

pema99:
Got a bit caught up in something over the weekend. I'll take a proper look at this soon 😅 My gut feel is that it can just land as is, and we'll fix up any minor gripes with folder structure etc.

Float3:
> AudioLinkSandbox is for stuff we don't want to distribute, so it shouldn't be in the package. It can go in the project, I think.

It is currently included in the "AudioLink Full.unitypackage", right?
Though for VPM purposes, it should probably be excluded from the package.

pema99:
It isn't in either unitypackage

pema99:
Waiting to sort out / investigate some remaining issues relating the install workflow

pema99:
Merging this now, and we fix remaining issues afterwards

PR #193, opened 2022-08-31
AudioLink Shader Compatability Utility

Float3:
A utility to upgrade absolute shader paths pointing to the old path to point to the new path

### DIFF
--- a/Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
+++ b/Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.IO;
 using UnityEditor;
 using UnityEngine;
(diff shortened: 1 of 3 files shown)


Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
```diff
@@ -0,0 +1,96 @@
...
+                {
+                    ReplaceInFile(path);
+                }
+            }
+        }
+
+        private static void UpgradeCgincFiles()
+        {
+            List<string> cgincs = FindCgincFiles(Application.dataPath);
+            cgincs.AddRange(FindCgincFiles(Path.GetDirectoryName(Application.dataPath) + "/Packages"));
```
pema99, line 55: What does Path.GetDirectoryName do when fed a directory path?

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
```diff
@@ -0,0 +1,96 @@
...
+            + "\n" + "\n" +
+            "(You can always run this utility manually via the " + MenuItemPath + " menu command)";
+
+        private const string DialogTitle = "AudioLink Shader Compatibility Utility";
+        private const string DialogOkButton = "I Made a Backup, Go Ahead!";
+        private const string DialogCancelButton = "No Thanks";
+
+
+        [MenuItem(MenuItemPath)]
+        public static void UpgradeShaders()
```
pema99, line 27: There should probably be an AssetDatabase.Refresh when we are done upgrading files

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        private static List<string> FindCgincFiles(string path)
+        {
+            List<string> cgincList = new List<string>();
+
+            string[] files = Directory.GetFiles(path);
+            foreach (string file in files)
+            {
+                if (file.EndsWith(".cginc"))
```
pema99, line 70: Maybe check if it ends with `.hlsl` as well. Some people use that for includes

pema99:
I like it

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
Float3: resolved in 7b153a899e8888fce0f0862f06d57349176e29c5

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
Float3: resolved in 7b153a899e8888fce0f0862f06d57349176e29c5

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
Float3: It returns the parent folder

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
Float3: are there any other valid extensions?

Float3: f16524a0669b033e7bbf5065dc4bae78732168cb

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            string[] shaderSource = File.ReadAllLines(path);
+            bool shouldWrite = false;
+            for (int i = 0; i < shaderSource.Length; i++)
+            {
+                string line = shaderSource[i];
+                if (line.Contains(OldAbsolutePath))
+                {
+                    line = line.Replace(OldAbsolutePath, NewAbsolutePath);
+                    while (line.Contains("/Packages"))
```
pema99, line 97: What is this

Float3: therefore we have to remove any prefixed /s

pema99: Is that really what this code is doing, though? The replace call would replace 'Packages' anywhere in the string, and Contains would check anywhere in the string. I'm assuming you got this right, I just wanna hear your thoughts.

Float3: I've tested this thoroughly, maybe you can re-read it.
this is just to make sure we don't have `#include \Packages\` because we essentially do `line.Replace("Assets","Packages")` so if it has a suffixed '\' it's then illegal

PR #199, opened 2022-09-14
Remove unused component from miniplayer screen

jaquadro:


Float3:
What component was removed this?

jaquadro:
Unused, disabled AVPro Video Screen component.  The active component is on the Output node instead.

Float3 (approved):

PR #202, opened 2022-10-13
Add YoutubePlayer editor-only component

rrazgriz:
Adds an editor-only component for piping a Youtube video link to a Unity video player, which can then output to the AudioLink source. This makes it easy to test with different audio without downloading individual audio files. It's additive on top of an audio source.

I originally built this for Avatar AudioLink testing, and it's my hope that it will be included in an easy-to-download AudioLink testing package for Avatars. It runs both with and without the VRC SDK and Udon, so it can run in Avatar, World, and non-VRC projects.

It's designed and #define'd to only run on Windows for now. Currently, it looks for a yt-dlp executable from the user's VRChat install folder. If not, it offers to download it to a hardcoded path in the project. This could be changed to a different path, and probably should for the package build.

Things that need to be added:
- Handling of the download with the package folder structure
- Addition to the Testing prefab for AudioLink Avatars
- In-editor handling of AudioLink MiniPlayer functionality

### Design

Uses code from the following sources:
UnityYoutubePlayer, courtesy iBicha (SPDX-License-Identifier: Unlicense) https://github.com/iBicha/UnityYoutubePlayer/
USharpVideo, Copyright (c) 2020 Merlin, (SPDX-License-Identifier: MIT) https://github.com/MerlinVR/USharpVideo/

Unity_MUGeewq77n

https://user-images.githubusercontent.com/47901762/195599861-e9961fde-3fd0-4d53-add7-fd44b86a8b6b.mp4


pema99:
Is this ready for review? You have the PR on draft mode.

float3:
From prior conversations, there were a couple things we wanted to explore first
like having this component on the AudioLinkAvatar prefab.
Also I don't think the component should be named YoutubePlayer since ytdlp is compatible with a wide variety of websites. I wouldn't be opposed to ytdlpPlayer or something like that.
Also I want to adjust this script so it can be used in standalone unity use cases (non-vrc).
I also wanted to try getting this to work with Internet radio or add a different component for that.

pema99:
Leave the radio thing for another pr. The other things sound reasonable

pema99:
Does this intentionally only touch the avatar prefab?

Packages/com.llealloo.audiolink/Runtime/AudioLinkAvatar.prefab
```diff
@@ -1,101 +1,120 @@
...
+  m_Script: {fileID: 11500000, guid: af9b9b5c52a71184bb2fad4c01f0bc4d, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  audioMaterial: {fileID: 2100000, guid: 5189d837989b9144e8e43d3ce618e450, type: 2}
-  audioMaterialInLeft: {fileID: 0}
-  audioMaterialInRight: {fileID: 0}
-  audioTextureExport: {fileID: 0}
-  audioData2d: {fileID: 2800000, guid: b07c8466531ac5e4e852f3e276e4baca, type: 3}
-  audioSource: {fileID: 8800328340110197116}
+  ytdlpURL: https://www.youtube.com/watch?v=SFTcZ1GXOCQ
```
pema99, line 81: lol

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -0,0 +1,303 @@
+using System;
```
pema99, line 1: Why is this class in the Runtime folder? Isn't it just for testing in Editor?

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -0,0 +1,303 @@
...
+            Process ytdlProcess = new Process();
+
+            ytdlProcess.StartInfo.WindowStyle = ProcessWindowStyle.Hidden;
+            ytdlProcess.StartInfo.CreateNoWindow = true;
+            ytdlProcess.StartInfo.UseShellExecute = false;
+            ytdlProcess.StartInfo.RedirectStandardOutput = true;
+            ytdlProcess.StartInfo.FileName = _ytdlpPath;
+            ytdlProcess.StartInfo.Arguments = $"--no-check-certificate --no-cache-dir --rm-cache-dir -f \"mp4[height<=?{resolution}]/best[height<=?{resolution}]\" --get-url \"{url}\"";
+
+            Debug.Log($"[AudioLink] Attempting to resolve URL '{url}'");
```
pema99, line 227: Is there really a reason to spam the log in the good path..

pema99: Silence is golden

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -0,0 +1,303 @@
...
+            }
+        }
+
+        /// <summary> Resolves a URL and set a videoplayer's URL to the resolved URL. </summary>
+        /// <param name="url">URL to resolve for playback</param>
+        /// <param name="resolution">Resolution (vertical) to request from yt-dlp</param>
+        /// <param name="player">VideoPlayer component to set URL on</param>
+        public static void ResolveAndSet(string url, int resolution, VideoPlayer player)
+        {
+            try
```
pema99, line 254: this try/catch is pointless, you already handle exceptions in Resolve

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -0,0 +1,303 @@
...
+        {
+            UpdateURL();
+            if (VideoPlayer.length > 0)
+                VideoPlayer.Play();
+        }
+
+        /// <summary> Set time to zero, resolve, and set URL </summary>
+        public void UpdateURL()
+        {
+            try
```
pema99, line 48: this try catch also seems pointless? can anything fail but the innermost Resolve?

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -0,0 +1,303 @@
...
+                return 0;
+        }
+
+        /// <summary> Set Video Player Playback Time (Seek) </summary>
+        /// <param name="time">Fraction of playback (0-1) to seek to</param>
+        public void SetPlaybackTime(float time)
+        {
+            if(VideoPlayer != null && VideoPlayer.length > 0)
+            {
+                if (!VideoPlayer.canSetTime)
```
pema99, line 75: If can **set** -> then **get**. I don't follow

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -0,0 +1,303 @@
...
+        public override void OnInspectorGUI()
+        {
+            base.OnInspectorGUI();
+            float playbackTime = 0;
+            bool hasPlayer = _ytdlpPlayer.VideoPlayer != null;
+            if(hasPlayer && _ytdlpPlayer.VideoPlayer.length > 0)
+                playbackTime = _ytdlpPlayer.GetPlaybackTime();
+            
+            EditorGUI.BeginDisabledGroup(!hasPlayer || !Application.IsPlaying(target) || !_ytdlpPlayer.VideoPlayer.isPlaying);
+            _reloadURL = EditorGUILayout.Toggle("Reload URL", _reloadURL);
```
pema99, line 124: Why a toggle and not a button? Am  I missing something?

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -0,0 +1,303 @@
...
+            {
+                webClient.DownloadFile(new Uri(_ytdlpDownloadURL), _localYtdlpPath);
+                Debug.Log($"[AudioLink] yt-dlp downloaded to '{_ytdlpPath}'");
+                AssetDatabase.Refresh();
+            }
+            catch(Exception e)
+            {
+                Debug.LogWarning($"[AudioLink] Failed to download yt-dlp from '{_ytdlpDownloadURL}' : " + e.Message);
+            }
+            webClient.Dispose();
```
pema99, line 297: Use a `using` statement instead of a manual dispose

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
float3: unity was complaining when it was in a folder called Editor

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -0,0 +1,303 @@
...
+                // Offer to download yt-dlp to the AudioLink folder
+                bool doDownload = EditorUtility.DisplayDialog("[AudioLink] Download yt-dlp?", "AudioLink could not locate yt-dlp in your VRChat folder.\nDownload to AudioLink folder instead?", "Download", "Cancel");
+                if(doDownload)
+                    DownloadYtdlp();
+
+                if(!Application.isPlaying)
+                    EditorApplication.ExitPlaymode();
+                
+                #elif UNITY_EDITOR_LINUX
+                    EditorUtility.DisplayDialog("[AudioLink] Missing yt-dlp", "try \"sudo pacman -S yt-dlp\"", "ok");
```
pema99, line 189: This only works on Arch based distros, and not even the most popular. If you are gonna put a help, at least include a command for Debian.

float3: I don't know debian commands and I don't wanna research them; can you help you?

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -0,0 +1,303 @@
...
+
+        private static string _ytdlpPath = "";
+        private static HashSet<Process> _runningYtdlProcesses = new HashSet<Process>();
+        private static bool _ytdlFound = false;
+
+        /// <summary> Initialize URL Resolver </summary>
+        static YtdlpURLResolver()
+        {
+            EditorApplication.playModeStateChanged += PlayModeChanged;
+            LocateYtdlp();
```
pema99, line 156: This is pretty cursed, I'm not gonna lie. You have no control over static initialization, and in Unity especially, static fields act in very funky ways. It also means that you will potentially get nagged by the script even if you never plan to use it. The class should probably just be a lazy singleton if you absolutely need global state.

More pressingly, you have no way to handle cleanup. You spin up processes and put them into this hashset, but if a domain reload happens, the hashset will be reset, and the processes will be left dangling, unless I am missing something.

I don't really get what the point of keeping the processes in this hashmap is, anyways. Could you enlighten me? Is there ever more than 1 in here?

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
pema99: `sudo apt install <package>`

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
pema99: For what reason? If you want it in here, you should at least wrap it in a `#if UNITY_EDITOR && !COMPILER_UDONSHARP`

pema99: But check. Maybe it's something simple.

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
pema99: argh, ytdlp is in a separate PPA, so you'd need to do:
```
sudo add-apt-repository ppa:yt-dlp/stable    # Add ppa repo to apt
sudo apt update                              # Update package list
sudo apt install yt-dlp                      # Install yt-dlp
```

pema99: So if you wanna keep this, just mention it is specific to Arch

rrazgriz: Resolved in 631f888e53bf119a63570ed455611d1e38ef151d

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
rrazgriz: Resolved in 631f888e53bf119a63570ed455611d1e38ef151d

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
rrazgriz: Old check, resolved in 017fd35fa743a75b5ee7b0914fd4a2328bfc13d4

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
rrazgriz: Removed, resolved in 017fd35fa743a75b5ee7b0914fd4a2328bfc13d4

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
rrazgriz: Removed, resolved in 017fd35fa743a75b5ee7b0914fd4a2328bfc13d4

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
rrazgriz: Removed, resolved in 631f888e53bf119a63570ed455611d1e38ef151d

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
rrazgriz: No special reasoning. Changed to button in aca2c1d6e00cff52cf03ecf550f8f453c7a63efa

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -45,15 +45,11 @@ public void UpdateAndPlay()
...
+            string resolved = YtdlpURLResolver.Resolve(ytdlpURL, 720);
+            if(resolved != null)
             {
+                VideoPlayer.url = resolved;
                 SetPlaybackTime(0.0f);
-                YtdlpURLResolver.ResolveAndSet(ytdlpURL, 720, VideoPlayer);
-            }
-            catch
-            {
-                VideoPlayer.Pause();
```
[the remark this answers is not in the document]
pema99: ie. VideoPlayer.Pause no longer being called in the error case

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -0,0 +1,283 @@
...
+
+            if (!File.Exists(_ytdlpPath))
+            {
+                #if UNITY_EDITOR_WIN
+                // Offer to download yt-dlp to the AudioLink folder
+                bool doDownload = EditorUtility.DisplayDialog("[AudioLink] Download yt-dlp?", "AudioLink could not locate yt-dlp in your VRChat folder.\nDownload to AudioLink folder instead?", "Download", "Cancel");
+                if(doDownload)
+                    DownloadYtdlp();
+
+                if(!Application.isPlaying)
```
pema99, line 184: Is this intentional? "If not playing -> stop playing" ?

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
float3: unity just doesn't like MonoBehaviours in Editor folders afaik, I wrapped it in `#if UNITY_EDITOR && !COMPILER_UDONSHARP` in ca8929b

pema99:
I went ahead and addressed the remaining gripes I had, ripping out that process stuff I was complaining about. When I was testing this, one annoyance did pop up: By default, the audio produced by the AudioLinkAvatar prefab will be very low when using this setup, because the Audio Source used is set to 0.01 volume. We support using any volume you want for the Audio Source in AudioLink now, so I suppose this should just be set 1.0 for the avatar prefab (since it is meant for testing anyways). Did you hit this @rrazgriz, or were you using it differently? I just drag in the prefab and enter playmode - works, but low volume

float3:
Good point, we should definitely set the avatar prefabs volume to 1.0, I had the same issue

rrazgriz:
Thanks for the changes, appreciate the help on the process handling side.

> Good point, we should definitely set the avatar prefab's volume to 1.0, I had the same issue

1.0 tends to be very loud on my system - I would go for 0.25-0.3 instead by default, subjectively. But an increase is warranted for sure, I always increase to 0.1-0.2.

float3:
@pema99 if we address the volume issue then this is merge-able right?

PR #203, opened 2022-10-14
Controller changes

float3:


float3:
not sure if `GetSettings()` should be called before or after the init values are set at

https://github.com/llealloo/vrc-udon-audio-link/pull/203/commits/34892654d7a17b68b4d7a0a2ee87609e947776a7#diff-e1e6d098d6acaa81cb7a1bca4d1557ba2440957d141d52be056090618984161aR99

float3:
> not sure if `GetSettings()` should be called before or after the init values are set at
> 
> [3489265#diff-e1e6d098d6acaa81cb7a1bca4d1557ba2440957d141d52be056090618984161aR99](https://github.com/llealloo/vrc-udon-audio-link/commit/34892654d7a17b68b4d7a0a2ee87609e947776a7#diff-e1e6d098d6acaa81cb7a1bca4d1557ba2440957d141d52be056090618984161aR99)

we decided that it should happen before

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveSurface.cs
```diff
@@ -7,13 +7,12 @@ namespace VRCAudioLink
...
     #if UDON
         using UdonSharp;
-        using VRC.Udon;
 
         public class AudioReactiveSurface : UdonSharpBehaviour
         {
             [Header("To use custom mesh, swap mesh in Mesh Filter component above")]
             [Header("AudioLink Settings")]
-            public UdonBehaviour audioLink;
+            public AudioLink audioLink;
```
pema99, line 15: doesn't look like this field is used at all anymore. I think we should remove it. Same goes for AudioReactiveSurfaceArray.cs

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
```diff
@@ -95,10 +92,12 @@ void Start()
                 }
                 else
                 {
-                    themeColorController.SetProgramVariable("audioLink", audioLink);
+                    themeColorController.audioLink = audioLink;
                     themeColorController.SendCustomEvent("UpdateAudioLinkThemeColors");
```
pema99, line 96: Does this still work now that the field has type `ThemeColorController`? This is basically equivalent to a method call, right?

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
```diff
@@ -95,10 +92,12 @@ void Start()
...
                 }
 
+                GetSettings();
```
pema99, line 99: The ordering here intends for the reset button to reset to the settings the user has chosen in the inspector, I assume?

pema99:
Assuming you've tested that the changes to the controller work as intended, this looks good to me. Thanks for fixing the linking up button

pema99 (approved):

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
float3: Correct

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
float3: [ffa9c3a](https://github.com/llealloo/vrc-udon-audio-link/pull/203/commits/ffa9c3ae9054efa9f68658f2392f442820b4267e)

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveSurface.cs
float3: [842a723](https://github.com/llealloo/vrc-udon-audio-link/pull/203/commits/842a723b4cf9d96308d39175fe5f75c0d14a3131)

PR #215, opened 2022-12-12
Rename ScriptingDefineManager to AudioLinkDefineManager and add logic…

float3:
… to set "AUDIOLINK_IMPORTED" keyword


float3:
Still needs testing

PR #230, opened 2023-03-12
Release 0.3.2 

float3:

PR #231, opened 2023-03-13
MenuItem

float3:


float3:
Do not merge this before #230 goes through

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkSceneManager.cs
```diff
@@ -0,0 +1,37 @@
...
+        [MenuItem("AudioLink/Add AudioLink Prefab to Scene", false)]
+        [MenuItem("GameObject/Add AudioLink Prefab to Scene", false, 49)]
+        static void AddAudioLinkToScene()
+        {
+#if VRC_SDK_VRCSDK3 && !UDONSHARP //VRC AVATAR
+            string[] paths = new string[]
+            {
+                "Packages/com.llealloo.audiolink/Runtime/AudioLinkAvatar.prefab"
+            };
+#else  //VRC WORLD or STANDALONE
```
pema99, line 18: don't add the controller in the case of standalone, it won't work anyways

pema99:
We should maybe consider folding this into AudioLinkImportFix.cs, potentially changing the name while doing so, just to keep this kind of stuff in one place. AudioLinkImportFix is currently a bit of a frankenstein of scene-and-package-management-things, which I actually think is fine, the name just isn't great.

float3:
> We should maybe consider folding this into AudioLinkImportFix.cs, potentially changing the name while doing so, just to keep this kind of stuff in one place. AudioLinkImportFix is currently a bit of a frankenstein of scene-and-package-management-things, which I actually think is fine, the name just isn't great.
> 
> (Just keep in mind if you move anything that the AUDIOLINK_STANDALONE define only works in that one file currently, added by our CI)

okay

float3:
@pema99 is this what you had in mind. I figured I would combine all three of these files since they are all pretty self contained

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkManager.cs
```diff
@@ -96,4 +100,66 @@ public static void InstallUdonSharp()
...
+            if (definesChanged)
+            {
+                string finalDefineString = string.Join(";", defineSet.ToArray());
+                PlayerSettings.SetScriptingDefineSymbolsForGroup(buildTarget, finalDefineString);
+                Debug.LogFormat("Set Scripting Define Symbols for selected build target ({0}) to: {1}",
+                    buildTarget.ToString(), finalDefineString);
+            }
+        }
+    }
+    public static class SceneManager
```
pema99, line 134: I was actually more talking about at the class level. I'd take this a step further and combine SceneManager and AudioLinkImportFix into one class, naming it appropriately. They both do scene-and-package (asset, basically) management, if you look at the existing code in this file.

pema99: The define manager is reasonably separate in purpose tho, so that split is fine, I think.

float3: what name do you propose?

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkDefineManager.cs
```diff
@@ -7,11 +7,12 @@
 namespace VRCAudioLink.Editor
 {
     [InitializeOnLoad]
-    public class ScriptingDefineManager
+    public class DefineManager
```
pema99, line 10: lets keep the pattern and calling it AudioLinkDefineManager

pema99 (approved):

PR #233, opened 2023-03-14
Clean up Ytdlp integration

pema99:
- NEVER prompt the user to download yt-dlp if we are not in the VRC usecase (VRC_SDK_VRCSDK3 is not defined)
- First, check if we can find yt-dlp in the users VRChat directory
- Next, check if we can find yt-dlp on the users PATH
- If none of that works, give up with a warning

Also, no longer force the user out of playmode when yt-dlp couldnt be found. Thats pretty annoying.

I found that the player doesn't actually work on Linux at all, because Unity on Linux doesn't support any of the required formats :( left the non-windows handling in since it might work on OSX.

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
@@ -156,7 +156,7 @@ namespace VRCAudioLink
             
             // Check for yt-dlp in VRC application data first
             _ytdlpPath = string.Join("\\", splitPath.Take(splitPath.Length - 2)) + "\\VRChat\\VRChat\\Tools\\yt-dlp.exe";
-            #elif UNITY_EDITOR_LINUX
+            #else
             _ytdlpPath = "/usr/bin/yt-dlp";
             #endif
             if (!File.Exists(_ytdlpPath)) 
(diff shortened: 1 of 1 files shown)


Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -156,7 +156,7 @@ public static void LocateYtdlp()
             
             // Check for yt-dlp in VRC application data first
             _ytdlpPath = string.Join("\\", splitPath.Take(splitPath.Length - 2)) + "\\VRChat\\VRChat\\Tools\\yt-dlp.exe";
-            #elif UNITY_EDITOR_LINUX
+            #else
```
float3, line 159: this path is now MAC or LINUX is that intentional?

float3 (approved):
lgtm, maybe we should add a warning if LINUX

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
float3: ah it looks like it is intentional

pema99: yes. its a decent first guess for both

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -167,7 +167,7 @@ public static void LocateYtdlp()
 
             // Offer to download yt-dlp to the AudioLink folder ONLY when we are not in VRC usecase,
             // since VRC's security team doesn't like us downloading executables.
-            #if !VRC_SDK_VRCSDK3
+            #if !VRC_SDK_VRCSDK3 && UNITY_EDITOR_WIN
```
float3, line 170: the cynic in me wants to find a Mac user to test this with

float3: but that should be a fraction of our userbase, so if you decide you don't care I'm fine with that

float3: I tested it: if you bake a disabled avatar the result will also be disabled

pema99: wha? did you leave this comment on the wrong PR or something. Whats this about baking

pema99: anyways, this doesn't worsen the situation on mac, so I think its fine.

PR #246, opened 2023-04-14
some changes

float3:


Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -16,14 +16,14 @@ namespace VRCAudioLink
...
-    using static UnityEngine.Shader;
+    using static Shader;
 
     using UnityEngine.Rendering;
 
     public class AudioLink : MonoBehaviour
 #endif
     {
-        const float AUDIOLINK_VERSION_NUMBER = 3.02f;
+        const float AudiolinkVersionNumber = 3.02f;
```
pema99, line 23: Audio**L**ink

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkEditor.cs
```diff
@@ -1,35 +1,66 @@
...
-        UdonSharpBehaviour[] allUdonSharpBehaviours = FindObjectsOfType<UdonSharpBehaviour>();
-        foreach (var behaviour in allUdonSharpBehaviours)
+        public override void OnInspectorGUI()
         {
-            if (behaviour.GetType().GetField("audioLink") != null)
+#if UDONSHARP
+            if (UdonSharpGUI.DrawDefaultUdonSharpBehaviourHeader(target)) return;
+#endif
+            // don't use Camera.main here because the main camera might not have the main camera tag?
+            if (FindObjectsOfType<Camera>().Length <= 1)
```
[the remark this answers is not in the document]
pema99: this is kinda wack to have in a hot-loop since the perf gets arbitrarily bad, but I guess it's not so bad considering it will only happen when the inspector is shown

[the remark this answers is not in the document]
float3: I tested this and couldn't see a slow down even with moderately sized scenes, but you are right, it's not ideal. we can probably make some assumptions and just use Camera.main

pema99:
Retarget this to master. Dev is kill.

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -100,6 +100,7 @@ public class AudioLink : MonoBehaviour
 
         [Header("Experimental (Limits performance)")]
         [Tooltip("Enable Udon audioData array. Required by AudioReactiveLight and AudioReactiveObject. Uses ReadPixels which carries a performance hit. For experimental use when performance is less of a concern")]
+        [Obsolete("Do not reference this field directly, use AudioLink.EnableReadback and AudioLink.DisableReadback instead")]
```
float3, line 103: if this field is manually accessed in editor or by an udon script then it will go out of sync with the camera state.
I see a couple of ways of solving this
- access modifiers such as internal (maybe even different access modifiers based on UNITY_EDITOR, although that is a little cursed)
- Obsolete attribute, though it seems like weak security
- Guarding the camera disabling logic with !UNITY_EDITOR

float3: I also added some buttons to the UI to toggle this state instead

float3: lmk what you think @pema99

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -304,11 +304,13 @@ void Start()
             gameObject.SetActive(true); // client disables extra cameras, so set it true
             transform.position = new Vector3(0f, 10000000f, 0f); // keep this in a far away place
 
+#if !UNITY_EDITOR
```
pema99, line 309: why?

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -494,10 +496,25 @@ void OnPostRender()
...
                 audioData = audioData2D.GetPixels();
+#endif
+            }
+        }
+        // ReSharper disable once UnusedMember.Global
+        public void OnAsyncGpuReadbackComplete(VRCAsyncGPUReadbackRequest request)
+        {
+            if (!request.hasError)
+            {
+                audioData = new Color[audioData2D.width * audioData2D.height];
```
pema99, line 513: allocation for no reason

Packages/com.llealloo.audiolink/Runtime/AudioLink.asmdef
```diff
@@ -3,17 +3,16 @@
     "references": [
         "UdonSharp.Runtime",
         "UdonSharp.Lib",
-        "VRC.Udon"
+        "VRC.Udon",
+        "VRC.Udon.Serialization.OdinSerializer"
```
pema99, line 7: why?

pema99:
IMO mixing the 2 first points with the rest is a cardinal sin, since it makes the diff way more cursed than it has to be. I don't want to be a dick here, but could you please at least split this into: refactoring (rename, dead code, reformat, style changes, etc.), bugfixes/usability (warnings, small behavior changes), readback changes (async readback + improves to readback UI, new texture format)

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
float3: if I do this then we can't reference the field ourselves

float3: or at least I don't know how

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -494,10 +496,25 @@ void OnPostRender()
         {
             if (audioDataToggle)
             {
+// only do this on android
```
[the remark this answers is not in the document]
float3: oops didn't mean to commit this

Packages/com.llealloo.audiolink/Runtime/AudioLink.asmdef
float3: referencing IUdonEventReceiver causes this otherwise
```
The type 'ISupportsPrefabSerialization' is defined in an assembly that is not referenced. You must add a reference to assembly 'VRC.Udon.Serialization.OdinSerializer, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
```

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
float3: in editor someone might toggle `audioDataToggle` in the inspector and wonder why it isn't working (because the cameras were turned off on start).

PR #251, opened 2023-04-26
first pass at ARB

float3:


pema99:
Pushed some fixes and improvements. Looking good so far.

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -495,6 +512,25 @@ void OnPostRender()
                 audioData = audioData2D.GetPixels();
             }
         }
+#else
+#if UDONSHARP
+        public void OnAsyncGpuReadbackComplete(VRCAsyncGPUReadbackRequest request)
```
pema99, line 517: can we just split this into 2 indentical named functions. The preprocessor spam is hard to read

pema99: #if UDONSHARP
public void OnAsyncGpuReadbackComplete(VRCAsyncGPUReadbackRequest request) { ... }
#else
public void OnAsyncGpuReadbackComplete(AsyncGPUReadbackRequest request) { ... }
#endif

float3: ok

PR #253, opened 2023-04-27
improve yt-dlp player

float3:


Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -242,29 +272,94 @@ public override void OnInspectorGUI()
...
+                                _ytdlpPlayer.VideoPlayer.Stop();
+                        }
+                    }
+                }
+
+                float volume;
+                using (new EditorGUI.DisabledScope(!_ytdlpPlayer.GetAudioSourceVolume(out volume)))
+                {
+                    EditorGUI.BeginChangeCheck();
+                    volume = EditorGUILayout.Slider(new GUIContent("  AudioSource Gain", EditorGUIUtility.IconContent("d_Profiler.Audio").image), volume, 0.0f, 1.0f);
```
teehzero: Labelling this "AudioSource Gain" in the UI led me originally to believe this would affect AudioLink internal gain only. A more canonical name like "Volume" might be better for UX.

float3:
closed

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
float3: this makes sense to me as it's also called Volume in the AudioSource component

PR #255, opened 2023-05-02
improve yt dlp player

float3:
continuation of #253


pema99:
It's a little wack that the diff is basically useless here because we renamed the file simultaneously with changing a bunch of code. Do you plan to address that or nah?

float3:
> It's a little wack that the diff is basically useless here because we renamed the file simultaneously with changing a bunch of code. Do you plan to address that or nah?

what do you suggest?

pema99:
I think if you rename the file back to its original name, git might understand the diff. Worth a try, you can always change it in a later commit

pema99:
So long as you have minimal changes in the file, git is made to understand when you are just renaming stuff, and it will keep the history

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -1,203 +1,176 @@
...
-#endif
+                AudioSource audioSourceOutput = videoPlayer.GetTargetAudioSource(0);
+                if (audioSourceOutput != null)
+                    audioSourceOutput.volume = Mathf.Clamp01(volume);
             }
         }
     }
 
-    public static class YtdlpURLResolver
+    public class Request
```
pema99, line 135: lets change the name of this to something more descriptive

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -1,203 +1,176 @@
...
+    public static class ytdlpURLResolver
     {
-        private static string _localYtdlpPath = Application.dataPath + "\\AudioLink\\yt-dlp.exe";
+        private static string _localytdlpPath = Application.dataPath + "\\AudioLink\\yt-dlp.exe";
 
         private static string _ytdlpPath = "";
         private static bool _ytdlpFound = false;
 
-        public static bool IsYtdlpAvailable()
+        public static bool IsytdlpAvailable()
```
pema99, line 148: Uppercase ytdlp here ? And in Locateytdlp?

Packages/com.llealloo.audiolink/Runtime/Scripts/YtdlpPlayer.cs
```diff
@@ -277,5 +259,160 @@ private static string LocateExecutable(string name)
...
+
+            bool hasVideoPlayer = _ytdlpPlayer.videoPlayer != null;
+            float playbackTime = hasVideoPlayer ? _ytdlpPlayer.GetPlaybackTime() : 0;
+            double videoLength = hasVideoPlayer ? _ytdlpPlayer.videoPlayer.length : 0;
+
+            using (new EditorGUI.DisabledScope(!available))
+            {
+                using (new EditorGUILayout.HorizontalScope())
+                {
+                    EditorGUILayout.LabelField(new GUIContent(" Video URL", EditorGUIUtility.IconContent("CloudConnect").image), GUILayout.Width(100));
```
pema99, line 300: " Video URL" space intentional?

pema99: looks like it

PR #258, opened 2023-05-20
Initialize Autogain Derate with positive value

teehzero:
Initializing Autogain Derate with a value of 0 results in significant harmonic distortion, discovered by playing a test track consisting of pure sine tones.

This change updates the AudioLinkAvatar prefab and AudioLink script to match the AudioLink world prefab.

~~I expect this change does not fix the core issue of Autogain's compensation initializing at such a high value in the shader, but this should at least hide it.~~

I originally tried to change AudioLink.shader:258 `float4 lastAutoGain = AudioLinkGetSelfPixelData(ALPASS_GENERALVU + int2(11, 0));` to `float4 lastAutoGain = max(0.01f, AudioLinkGetSelfPixelData(ALPASS_GENERALVU + int2(11, 0)));`. When that didn't immediately work I just poked around at stuff until I found something I could work with. Turns out you can configure an initial texture right on the render texture itself.


pema99:
I don't understand why you are adding a new texture. Isn't this change just about changing the default value for derate? Why the init texture?

pema99:
Mirroring from discord:

Changing the init texture really doesn't seem like the right way to fix this to me

We could just not adjust gain if the last value is 0 or below some epsilon, like adding it to the branch condition for example

Unless  I'm missing something

pema99:
It's a bit hard for me to say with certainty without fiddling with it myself. I'll do so shortly unless 3 does

float3:
You go ahead if you want, I'm busy today

teehzero:
> Mirroring from discord:
> 
> Changing the init texture really doesn't seem like the right way to fix this to me
> 
> We could just not adjust gain if the last value is 0 or below some epsilon, like adding it to the branch condition for example
> 
> Unless I'm missing something

The init texture I uploaded changes the value for only a single pixel. I think changing it is fine since there should be no other circumstances in regular use where both lastAutogain and _AutogainDerate will reach zero. Initializing in this way does not interfere with the regular operation of any other AudioLink components, and it does not require any additional infrastructure in the code to handle. There may be a better value to initialize to, but after everything has smoothed out we shouldn't need to worry about handling 0 cases.

PR #261, opened 2023-06-10
Implement Media State API

fundale:
Used for sending Video / Audio Player state to Avatar / World props
Has Player Volume, Player Time, Player Playback state, Player loop / playlist state


float3:
Hi, thanks for the PR, just fyi we increment the version number on each release and we decided the next release will be 1.0.0 as we're switching to semver

float3:
currently I'm keeping track of changes on the changelog branch, but I realize that's suboptimal so I'll try to merge that soon. other than that I think the feature you implemented is fine, we will have to test and get back to you with any change requests

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -333,6 +373,48 @@ void Start()
...
+        None       0 (0.00)
+        Loop       1 (0.10)
+        LoopOne    2 (0.20)
+        Random     3 (0.30)
+        RandomLoop 4 (0.40)
+        */
+        public void SetMediaLoop(MediaLoop loopstate) {
+
+            int loop = (int)loopstate;
+            audioMaterial.SetFloat("_MediaLoop", (float)loop / 10);
```
pema99, line 414: Why the division by 10? Values >1 should work just fine

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -21,10 +41,30 @@ public class AudioLink : UdonSharpBehaviour
 
     using UnityEngine.Rendering;
 
+        public enum MediaPlaying
```
pema99, line 44: These enums really shouldn't be duplicated. You should move them such that they are compiled into both code paths.

pema99: Lmk if you need help with that

pema99:
Thanks. The changes look mostly fine, but all the stuff relating to changing the version number should be removed.

Docs/PlayerAPI.md
```diff
@@ -0,0 +1,82 @@
...
+            audiolink.SetMediaPlaying(MediaPlaying.Error);
+        }
+
+    }
+
+    //Update Media Volume
+    public void UpdateVolume(float volume) {
+
+        //Ranges from 0 to 1
+        if (Utilities.IsValid(audiolink)) audiolink.SetMediaVolume(volume);
```
techanon, line 50: You should use the fully-qualified namespace for the `VRC.SDKBase.Utilities.IsValid` to avoid naming collisions with other assets. Many people have come across this issue historically.

float3:
@fundale this pr is set so maintainers can edit your branch, I'm gonna assume this is intentional and will make the adjustments we need to merge it this weekend

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
fundale: I wasn't sure that it'd handle more than 1

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
fundale: Is there some other way I can do that?
When I tried it either complained about the enum coming before the using statements
Or complained about being embedded in a class

fundale:
> Thanks. The changes look mostly fine, but all the stuff relating to changing the version number should be removed, for the reasons @float3 points out

Assuming Maintainers can discard those changes right?
Or do I have to revert them?

techanon:
With the removal of the /10, don't forget to update the docs/comments to reflect that change.

float3:
> > Thanks. The changes look mostly fine, but all the stuff relating to changing the version number should be removed, for the reasons @float3 points out
> 
> Assuming Maintainers can discard those changes right?
> Or do I have to revert them?

I'll change it

float3:
@fundale 
I'm getting 
```
error: Authentication error: Authentication required: You must have push access to verify locks
error: failed to push some refs to 'https://github.com/fundale/vrc-udon-audio-link.git'
```
for some reason I can't push to your fork

here are the changes we require in a patch: https://gist.github.com/float3/cd364f25cd3f71cce010a94fd4863bee

float3 (approved):

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -333,6 +373,48 @@ void Start()
             }
         }
 
+        //Set Media Volume display
+        public void SetMediaVolume(float volume) {
```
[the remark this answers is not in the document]
float3: for the future we have format.sh file you can run to format the code

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
float3: this still needs to be addressed

PR #274, opened 2023-08-13
Cached local player

Nestorboy:
* Cached the Networking.LocalPlayer.
* Made player validation use Utilities.IsValid instead of null checking.


float3 (approved):
lgtm

PR #285, opened 2023-09-25
Add Midi Control Script

fundale:
Added a Midi Control script for manually controlling AudioLink config,
Might need reworked for new Controller UI,
Probably needs a Controller Locking mechanism first...


pema99:
Thanks for the PR. Probably going to hold off on this one for a while since we are currently trying wrap up the upcoming release asap

fundale:
That was unintended...

PR #286, opened 2023-10-02
Add Midi Controller (again)

fundale:
Same as before just resubmitting pr, as I broke the last one...

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMidiControl.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMidiControl.cs
@@ -1,0 +1,337 @@
+﻿
+using UdonSharp;
+using UnityEngine;
+using VRC.SDKBase;
+using VRC.Udon;
+
+namespace AudioLink
+{
+
+    [UdonBehaviourSyncMode(BehaviourSyncMode.None)]
+    public class AudioLinkMidiControl : UdonSharpBehaviour
+    {
+        public AudioLinkController _AudioLinkController;
+        public bool allowMidi;
+        public int midiChannel;
+
+        public bool _debugMidi;
+
+        void Start()
+        {
+            
+        }
+
+        public override void MidiNoteOn(int channel, int key, int value) {
+
+            
+
+            // Check On Midi
+            if (allowMidi && channel == midiChannel && Networking.IsOwner(_AudioLinkController.gameObject) && VRC.SDKBase.Utilities.IsValid(_AudioLinkController)) {
+
+                if (VRC.SDKBase.Utilities.IsValid(_AudioLinkController.themeColorController)) {
+
+                    ThemeColorController _ThemeColorController = _AudioLinkController.themeColorController;
+                
+                    switch (key) {
+
+                        case (int)MidiIds.ColorChord:
+                        // Enable Color Chord
+                        _ThemeColorController.themeColorDropdown.value = 0;
+                        break;
+
+                        case (int)MidiIds.Reset:
+                        // Trigger Controller Reset
+                        _AudioLinkController.ResetSettings();
+                        _ThemeColorController.ResetThemeColors();
+                        break;
+
+                        default:
+                        // No-op
+                        break;
+
+                    }
+
+                }
+
+            }
+
+            if (_debugMidi) {
+
+                //Debug.Log(channel);
+                Debug.Log(key);
+                Debug.Log(value);
+
+            }
+
+        }
+
+        public override void MidiNoteOff(int channel, int key, int value) {
+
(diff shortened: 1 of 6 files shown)


Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMidiControl.cs
```diff
@@ -0,0 +1,337 @@
+﻿
+using UdonSharp;
```
pema99, line 2: This whole thing should have a `#if UDONSHARP` around it, so we don't break the standalone (non-vrc) usecase

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMidiControl.cs
```diff
@@ -0,0 +1,337 @@
...
+    [UdonBehaviourSyncMode(BehaviourSyncMode.None)]
+    public class AudioLinkMidiControl : UdonSharpBehaviour
+    {
+        public AudioLinkController _AudioLinkController;
+        public bool allowMidi;
+        public int midiChannel;
+
+        public bool _debugMidi;
+
+        void Start()
```
pema99, line 19: empty start?

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMidiControl.cs
```diff
@@ -0,0 +1,337 @@
...
+using VRC.SDKBase;
+using VRC.Udon;
+
+namespace AudioLink
+{
+
+    [UdonBehaviourSyncMode(BehaviourSyncMode.None)]
+    public class AudioLinkMidiControl : UdonSharpBehaviour
+    {
+        public AudioLinkController _AudioLinkController;
```
pema99, line 13: why the mixed naming conventions? _PasalCase, _camelCase and camelCase

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMidiControl.cs
```diff
@@ -0,0 +1,337 @@
...
+
+                if (VRC.SDKBase.Utilities.IsValid(_AudioLinkController.themeColorController)) {
+
+                    ThemeColorController _ThemeColorController = _AudioLinkController.themeColorController;
+                    
+                    switch (key) {
+
+                        case (int)MidiIds.ColorChord:
+                        // Disable Color Chord
+                        _ThemeColorController.themeColorDropdown.value = 1;
```
pema99, line 81: This will only work with the old controller. Note how ThemeColorController.cs does not define themeColorDropdown anywhere, because it isn't a dropdown anymore

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMidiControl.cs
```diff
@@ -0,0 +1,337 @@
...
+using VRC.Udon;
+
+namespace AudioLink
+{
+
+    [UdonBehaviourSyncMode(BehaviourSyncMode.None)]
+    public class AudioLinkMidiControl : UdonSharpBehaviour
+    {
+        public AudioLinkController _AudioLinkController;
+        public bool allowMidi;
```
pema99, line 14: what's the point of having a boolean to disable the behavior? You can just disable the component itself, this seems redundant to me.

pema99:
I think this needs some revisiting to be compatible with the new controller, or both if you are interested in that. We hard-forked the old controller when making the new one.

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMidiControl.cs
fundale: Setting the component's .enabled state doesn't prevent the "VRC Midi Listener" from calling it's functions, so a allowMidi bool is required to allow for toggling it.

pema99:
I just tested this on my own midi keyboard (Novation Launchkey Mini). The controls were basically unusable. I'm curious, which Midi controller are you using? Perhaps this needs re-mappable controls to really be useful

fundale:
> I just tested this on my own midi keyboard (Novation Launchkey Mini). The controls were basically unusable. I'm curious, which Midi controller are you using? Perhaps this needs re-mappable controls to really be useful

Unfortunately I don't have any Physical Midi device, I used Loop-Midi for testing, hence why the inputs are defined in enum.
thinking now might look into defining the enum values in a swappable script

fundale:
TODO: Make scene for easily mapping new layouts

float3:
marked this as draft, let us know if you want a re-review

PR #288, opened 2023-10-15
Update AddAudioLinkToScene

techanon:
Update AddAudioLinkToScene to check for prefab existence before adding a new one to the scene.
Tested both VRCWorld (covers logic for VRCAvatar and Standalone) and CVRCCK.

### DIFF
--- a/Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkAssetManager.cs
+++ b/Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkAssetManager.cs
@@ -1,4 +1,6 @@
 ﻿using System.IO;
+using System.Linq;
+using System.Collections.Generic;
 using UnityEditor;
 using UnityEditor.SceneManagement;
 using UnityEngine;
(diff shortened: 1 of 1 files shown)


Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkAssetManager.cs
```diff
@@ -37,7 +40,7 @@ private static void ReimportPackage()
         }
 
 #if !AUDIOLINK_STANDALONE
-        [MenuItem("Tools/AudioLink/Open AudioLink Example Scene")]
+        [MenuItem("AudioLink/Open AudioLink Example Scene")]
```
pema99, line 43: looks like you have an unintended change here after your last pr went to master?

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkAssetManager.cs
```diff
@@ -51,50 +54,80 @@ public static void OpenExampleScene()
...
+                instance = (GameObject)PrefabUtility.InstantiatePrefab(sourceAsset);
+                Undo.RegisterCreatedObjectUndo(instance, "Undo create prefab instance");
             }
+
+            return instance;
+        }
+
+        private static T[] GetComponentsInScene<T>(bool includeInactive = true) where T : Component
+        {
+            var stage = PrefabStageUtility.GetCurrentPrefabStage();
```
pema99, line 117: can you TL;DR me why you need this?

pema99:
looks good overall, just that minor issue I commented about above

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkAssetManager.cs
techanon: It's just a convenience method I pulled from my own scripts which finds the list of all components that exist within the current scene (or prefab preview stage if that's active). This way it can easily find AudioLink in the scene regardless of where the user parents it to without relying on gameobject names, which is a fragile search method.

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkAssetManager.cs
techanon: Ah yea. Fixed.

pema99:
lgtm

PR #295, opened 2024-02-07
Fixed an error with null reference on VRC exit.

ShingenPizza:
I noticed that whenever* i exit VRC while on my map with AudioLink, while i have that map's project open**, i get an error in the console:
image

Apparently during cleaning stuff up the `DisableAudioLink()` method gets called after `audioRenderTexture` becomes null, so an attermpt to set its `updateMode` throws an exception.

This PR fixes it by checking if `audioRenderTexture` is null.

\* sometimes it stops doing that, but works again after project reopen
\** no matter if VRC has been run through the "build and test" or normally through Steam

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
@@ -738,7 +738,7 @@ namespace AudioLink
         public void DisableAudioLink()
         {
             _audioLinkEnabled = false;
-            audioRenderTexture.updateMode = CustomRenderTextureUpdateMode.OnDemand;
+            if (audioRenderTexture != null) { audioRenderTexture.updateMode = CustomRenderTextureUpdateMode.OnDemand; }
 #if UDONSHARP
             SetGlobalTexture(_AudioTexture, null);
 #else


float3 (commented):
I don't think this check is needed because in no sane state of audiolink will audioRenderTexture be null, but since this isn't in a hot loop or in a frequently called method it's fine to add

float3:
@pema99 can merge?

pema99:
what.. I'd like to understand why this happens

> I noticed that whenever* i exit VRC while on my map with AudioLink, while i have that map's project open**, i get an error in the console:

doesn't udon have some kind debugging mode when you do this?

pema99:
it looks ok to merge, but I guess if we do this we should match it with a null check in EnableAudioLink

ShingenPizza:
> what.. I'd like to understand why this happens

assets packaged with that scene somehow get unloaded faster than the scene itself, so the references become invalidated? idk

> it looks ok to merge, but I guess if we do this we should match it with a null check in EnableAudioLink

it surely wouldn't hurt,
but it is probably impossible for it to be null during the enable action (unless someone starts playing around with it too much),
and if it somehow ended up as null during the enable, then the whole AL probably should stop working altogether with some major error message, because it would break soon after anyway...

pema99:
fair point. this seriously should never happen, so I'm just going to assume its a bug with the udon debugging setup

ShingenPizza:
> doesn't udon have some kind debugging mode when you do this?

>fair point. this seriously should never happen, so I'm just going to assume its a bug with the udon debugging setup

to be clear, as i wrote before, it also happens when you start it up from Steam.
i do have all the debug [launch options](https://docs.vrchat.com/docs/launch-options) turned on, but i'm pretty sure it doesn't affect VRC's behaviour, just increases the logging.
i don't know how does the error-appearing-in-my-Unity-editor-console work, but it wouldn't be the first time.
normally you just can't see errors such as this, and it doesn't really matter that it does happen when you're shutting down the whole process, however it can still be seen in logs. at least with my increased logging launch options.
i've also just had the same bug happen on my friend's map, and somehow it appeared in my homeworld's project's console too.

i'd call it a bug with the order of things that get cleaned up while shutting down VRC, possibly due to some race conditions.

thanks for merging.

PR #296, opened 2024-02-10
Format

float3:


pema99:
Can we stick to non-egyptian bracing? This is the most common style for Unity C# code

pema99 (approved):

Issue #278, opened 2023-09-10
Please Document the VPM Install Method


owlboy:
**Feature Description:**
Please document the install method for those who use the CLI / VPM.

`vpm add package com.llealloo.audiolink`

**Additional context:**
Not everyone develops on Windows! Thanks!

float3:
you can make a pr

owlboy:
The reason I'm looking for the docs on this is for an authoritative answer. I'm not the authority. 😅

pema99:
Sorry never gotten around to this since I've never used the vpm CLI myself

float3:
> Sorry never gotten around to this since I've never used the vpm CLI myself

I'll write it

float3:
https://github.com/llealloo/vrc-udon-audio-link/tree/cli-instructions?tab=readme-ov-file#linuxmacoswindows

@owlboy here's my draft, If you wish to give feedback

float3:
merged in https://github.com/llealloo/vrc-udon-audio-link/pull/297

PR #297, opened 2024-02-10
document cli installation method

float3:


float3:
conversation at https://github.com/llealloo/vrc-udon-audio-link/issues/278

PR #303, opened 2024-03-25
Improve handling of search for YTDL executables

techanon:
- Improve path search logic to include some standard YTDL (not just yt-dlp) executable names.  
- Add menu option for specifying a precise custom executable location under Tools -> AudioLink -> Select - Custom YTDL Location.  
- Move _localytdlpPath use to windows only as it's an exe specific path.  
- Update LocateExecutable to support intaking a list of possible known ytdl executable names to improve search reliability.  
- Update LocateExecutable to support PATH quirks with certain MSeries Macbooks.

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
@@ -147,6 +147,12 @@ namespace AudioLink
         [MenuItem(userDefinedYTDLPathMenu, priority = 1)]
         private static void SelectYtdlInstall()
         {
+            if (Menu.GetChecked(userDefinedYTDLPathMenu))
+            {
+                EditorPrefs.SetString(userDefinedYTDLPathKey, string.Empty);
+                return;
+            }
+
             string ytdlPath = EditorPrefs.GetString(userDefinedYTDLPathKey, "");
             string tpath = ytdlPath.Substring(0, ytdlPath.LastIndexOf("/", StringComparison.Ordinal) + 1);
             string path = EditorUtility.OpenFilePanel("Select YTDL Location", tpath, "");
@@ -182,6 +188,7 @@ namespace AudioLink
                     Debug.Log($"[AudioLink:ytdlp] Custom YTDL location found: {customPath}");
                     _ytdlpPath = customPath;
                     _ytdlpFound = true;
+                    return;
                 }
 
                 Debug.LogWarning($"[AudioLink:ytdlp] Custom YTDL location detected but does not exist: {customPath}");
@@ -343,11 +350,7 @@ namespace AudioLink
                     EditorGUI.BeginChangeCheck();
                     _ytdlpPlayer.ytdlpURL = EditorGUILayout.TextField(_ytdlpPlayer.ytdlpURL);
                     if (EditorGUI.EndChangeCheck())
-                    {
                         EditorUtility.SetDirty(_ytdlpPlayer);
-                    }
-
-                    ;
                     _ytdlpPlayer.resolution = (ytdlpPlayer.Resolution)EditorGUILayout.EnumPopup(_ytdlpPlayer.resolution, GUILayout.Width(65));
                 }
 
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
@@ -19,13 +19,11 @@ namespace AudioLink
         public string ytdlpURL = "https://www.youtube.com/watch?v=SFTcZ1GXOCQ";
         ytdlpRequest _currentRequest = null;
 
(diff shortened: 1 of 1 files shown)


float3:
under what scenarios?

float3:
I don't remember what format ytdlp serves when the source is youtube, but maybe there's a way to get it to serve a specific format? if that was the case I would be okay with having it available on linux

float3:
I'm not sure if we want to support ytdl

float3:
> So it's an editor specific thing then. Ok.

did you test your changes on mac in editor?

techanon:
> did you test your changes on mac in editor?

That specific Mac part is ported from the changes I made to VideoPlayerShim based on the feedback from an M-Series Mac user. The shim uses somewhat different search logic, but the path requirement for m-series is the same.

> I'm not sure if we want to support ytdl

Is there a reason why not? Also it's all local side. Isn't it better to support the canonical name of ytdl as well as the ytdlp fork when searching for a users local install? The search logic prioritizes VRChat's official install first anyways, as well as ytdlp being searched for first in the list of possible names.

Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
```diff
@@ -155,36 +172,43 @@ public static bool IsytdlpAvailable()
...
+            if (!string.IsNullOrEmpty(customPath))
+            {
+                if (File.Exists(customPath))
+                {
+                    Debug.Log($"[AudioLink:ytdlp] Custom YTDL location found: {customPath}");
+                    _ytdlpPath = customPath;
+                    _ytdlpFound = true;
+                }
+
+                Debug.LogWarning($"[AudioLink:ytdlp] Custom YTDL location detected but does not exist: {customPath}");
```
pema99, line 187: This doesn't seem right? There should be no warning printed if we found the executable in the branch above, right?

Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
```diff
@@ -143,6 +141,25 @@ public static class ytdlpURLResolver
...
         private static bool _ytdlpFound = false;
 
+        private const string userDefinedYTDLPathKey = "YTDL-PATH-CUSTOM";
+        private const string userDefinedYTDLPathMenu = "Tools/AudioLink/Select Custom YTDL Location";
+
+        [MenuItem(userDefinedYTDLPathMenu, priority = 1)]
+        private static void SelectYtdlInstall()
+        {
+            string ytdlPath = EditorPrefs.GetString(userDefinedYTDLPathKey, "");
+            string tpath = ytdlPath.Substring(0, ytdlPath.LastIndexOf("/", StringComparison.Ordinal) + 1);
```
pema99, line 151: Won't this break if you use `\` as the path separator?

Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
```diff
@@ -301,7 +345,9 @@ public override void OnInspectorGUI()
                     if (EditorGUI.EndChangeCheck())
                     {
                         EditorUtility.SetDirty(_ytdlpPlayer);
-                    };
+                    }
+
+                    ;
```
pema99, line 350: typo looks like

pema99:
Looks sane overall. Perhaps there should be a way to clear the custom path you've set previously though?

Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
techanon: Return statement is missing. Will fix.

Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
techanon: I did try using the `\` separator check but what I saw was that Unity was normalizing to `/` for the paths.

techanon:
The current way that the custom path gets cleared is by clicking the menu option to open the file select dialog, and then clicking cancel. This returns an empty string path which is detected as not provided, thus falling back on the auto-detect methods.

float3:
> The current way that the custom path gets cleared is by clicking the menu option to open the file select dialog, and then clicking cancel. This returns an empty string path which is detected as not provided, thus falling back on the auto-detect methods.

yeah this is not an acceptable method of clearing the custom path

techanon:
We can switch it to a menu toggle style. When active, selecting it again implicitly clears it, then to pick a new path, select the menu again.
How's that sound?

pema99:
Sounds like an improvement to me

techanon:
Meant to push to a separate branch.

pema99:
Got some conflicts now, probably with your other PR

techanon:
@pema99 fixed.

PR #304, opened 2024-03-25
Add _Udon_VideoTex global texture support for testing Avatars

techanon:
Add support for exporting the YTDLP video player texture to the standard _Udon_VideoTex global texture property.  
The texture export logic is baked into the YtdlpPlayer component.


techanon:
For... avatars? Specifically playmode testing. That's the only place that the YtdlpPlayer component runs on afaik.

float3:
I'll talk to pema and lea to see if this is something we want to have in audiolink

techanon:
Some additional reasoning for posterity:
- The _Udon_VideoTex is an upcoming global shader standard for exposing video player textures to avatars.
- Custom and Uber shaders (like [poiyomi](https://twitter.com/poiyomi/status/1765463629942387186?s=46&t=UuLUxGsI88ojFZRoJsW3fQ)) are beginning to integrate the standard and it will increase in use.
- The AudioLinkAvatar prefab already comes with a video player for testing external media. Including this feature will help adoption and reduce creator friction by enabling testing it in playmode alongside the already existing audio reactive testing.
- Efforts to aid adoption by other VideoPlayers (both standalone and prefab) aside from ProTV is starting underway.

float3:
@pema99 can you check this out

Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
```diff
@@ -415,7 +465,17 @@ public override void OnInspectorGUI()
...
             }
         }
-    }
 
+        /// <summary>
+        /// Draws some text in the same style as the [Header] attribute.
+        /// </summary>
+        /// <param name="header">the desired text to draw</param>
+        /// <param name="fontSizeDelta"></param>
+        public static void Header(string header, int fontSizeDelta = 0)
```
pema99, line 474: this doesn't seem like something that needs to be made public

pema99:
The changes look fine. And I'm fine with the change in principle... but I'm wondering how we can make it clear that this isn't an AudioLink feature, just a convenience provided to the user. Perhaps a link in the UI to some info about it? What do you think @float3

Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
techanon: That can change. This was just some helper methods I pulled over from a utility class.

techanon:
image
How's this look for the notice?

float3:
> The diff looks fine. And I'm fine with the change in principle... but I'm wondering how we can make it clear that this isn't an AudioLink feature, just a convenience provided to the user. Perhaps a link in the UI to some info about it? What do you think @float3 

ideally I would prefer to not have this in the codebase

techanon:
I still believe that the accessibility of including this with Audiolink for testing dynamic external content out-weighs the concern for user confusion, especially if proper clear messaging is included as such.

Do you have any alternate ideas on easy implementation into a user's avatar testing flow?

pema99:
From a principled point of view, I agree with @float3. From a pragmatic point of view, I agree with @techanon. Perhaps we should hear some of the other frequent contributors out about it.

fundale:
> image How's this look for the notice?

May help catch the users attention to the notice better if it were a warning

techanon:
I don't think escalating it to a warning is necessary here.  
Usually a warning signifies some sort of optional action a user can take.  
This is simply an informational notice about the context of the feature.  
Warnings would confuse more users than it would help imo.

techanon:
I have added _ST control for testing the tiling/offset options.
Demo https://streamable.com/x2v0iw

techanon:
Don't merge this yet. Found a math defect.

techanon:
Ok, ready for review.

pema99:
Conflicts here as well 😅

techanon:
@pema99 fixed.

pema99:
I changed the UI to be slightly more muted. Also, I'm making the default state disabled. It seems like an OK compromise considering several maintainers were on the fence about including this.

image

PR #305, opened 2024-04-16
Redundant Media Start Fix

techanon:
Fix the video player restarting the media ~3 seconds in by offloading the video player Play action to the `prepareCompleted` event.

Swap to using SerializedProperty for the url and resolution fields.
Simplify a couple ternary conditions.


float3 (approved):
lgtm

Issue #306, opened 2024-04-24
Suggestion: Reduce Texture Size For Controller Icons


JLChnToZ:
**Feature Description:**
As I observed, every time I upload a world contains a AudioLink controller, it includes controller icon textures but in unreasonably high resolution (1024x1024 each). I don't think it worth to be that large, so I suggests it can be convert to SDF/MSDF format with a much lower resolution to reduce build size. P.S. if I have those source (vector) files, I can help convert them.

**Additional context:**
![image](https://github.com/llealloo/audiolink/assets/4715281/88c7099d-57d0-4e99-8dc1-073bcc21ab3e)
![image](https://github.com/llealloo/audiolink/assets/4715281/a5df0bed-adf5-4f40-9a9d-6310fb6fca3e)
![image](https://github.com/llealloo/audiolink/assets/4715281/2edf78af-f01f-4a44-8854-a01b523afca7)
![image](https://github.com/llealloo/audiolink/assets/4715281/3b1fa155-5638-4244-a722-8efb3586b80c)

There are several more textures/icons I think they are too large as well, but I think it is enough to show above as examples.

float3:
good point, we should do this

float3:
I think clamping import size for now is a good start and then we should look into need, do we want to build msdf baking infrastructure inside of unity for futureproofing?

JLChnToZ:
As I know MSDF baking requires external executable, IDK is it good to include the whole pipeline within AudioLink package.

pema99:
If you are going to bake SDF textures for the icons I would just do it once, don't have to bother making baking infrastructure. I'm not personally familiar with tools for doing this, but if either of you would like to give it a shot I can provide guidance on how the controller shader is written (though I imagine the shader changes should be fairly straight forward?).

@llealloo might still have the source files lying around

JLChnToZ:
I do have experience on doing this, I did it for all icons in my own video player.

Once I got the source files I could immediately work on this. 😉

float3:
> As I know MSDF baking requires external executable, IDK is it good to include the whole pipeline within AudioLink package.

not in the package but in Audiolinkdev unity project in the repo, which isn't shipped to users

PR #308, opened 2024-05-07
Convert AudioLink panel icons & shader to use MSDF format

JLChnToZ:
This is the changes according to #306.
If any more icons that need to migrate, I will include in here as well.


float3:
looks good to me, can you add the source files as well?

JLChnToZ:
Yes I can, but IDK where to put it.

pema99:
Put them somewhere in the asset folder of AudioLinkSandboxProject. Doesn't really matter where

pema99 (approved):

pema99:
Thanks 👍

PR #309, opened 2024-05-08
Change default Ytdlp URL to something more tolerate

BluWizard10:
I'm probably going to be very salty about this, but I think the default music URL is just really hard to listen to. Especially if I have to listen to it when helping other people out in a sharescreen when teaching people how to test AudioLink.

So, I've changed it. **It's the same exact video that is played in llealloo's AudioLink V2 World found [here](https://www.youtube.com/watch?v=dZPcLi2-5dE).** Many of the songs in this link are copyright-free and hit on all 4 bands, so this shouldn't be a problem.

Thank you for your consideration.

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/AudioLinkAvatar.prefab
+++ b/Packages/com.llealloo.audiolink/Runtime/AudioLinkAvatar.prefab
@@ -29,7 +29,7 @@ Transform:
   m_LocalRotation: {x: 0, y: 0, z: 0, w: 1}
   m_LocalPosition: {x: 0, y: 0, z: 0}
   m_LocalScale: {x: 1, y: 1, z: 1}
-  m_ConstrainProportionsScale: 1
+  m_ConstrainProportionsScale: 0
   m_Children: []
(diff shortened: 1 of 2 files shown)


orels1:
The huge upside of the current track is that it has very clear bass hits from the very beginning, you don't need to wait for the track to ramp up or anything like that.

I think its much better for actual effects work than the most of the ones provided in the world / via your link, as it gives very clear bands with a lot of space between hits right from the start, so you can quickly work on your shaders/materials right beside it.

You can just mute your game view / mute the stream audio in discord if that's bothering you so much :/

float3:
orels makes a good point, the default track, in editor, should be carefully chosen to accommodate some common usecases

suggestions: we can have a poll or have some member of the community make a track for us

pema99:
The longer video also may take longer to load? I'm not sure, it's a guess

BluWizard10:
> The longer video also may take longer to load? I'm not sure, it's a guess 

It loads just as the same amount of time it took for me, but I could be wrong.

Either way, I may consider elevating this to a GitHub Issue instead of a PR.

PR #311, opened 2024-05-12
Adjustable Sync Modes

fundale:
lmk if I missed anything

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
@@ -125,8 +125,6 @@ namespace AudioLink
         {
             
             SetTransformSync(gainSlider.transform, syncMode, 2);
-            //SetTransformSync(trebleSlider.transform, syncMode, 2);
-            //SetTransformSync(bassSlider.transform, syncMode, 2);
             SetTransformSync(fadeLengthSlider.transform, syncMode, 3);
             SetTransformSync(fadeExpFalloffSlider.transform, syncMode, 3);
 
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
@@ -141,6 +141,8 @@ namespace AudioLink
             UpdateSyncMode(autoGainToggle.transform, syncMode, ControllerSyncMode.None);
             UpdateSyncMode(powerToggle.transform, syncMode, ControllerSyncMode.ExcludePower);
 
+            themeColorController.networkSynced = (int)syncMode < (int)ControllerSyncMode.None;
+
         }
 
         void Start()
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
@@ -110,36 +110,36 @@ namespace AudioLink
             return controllerTransform.GetComponent<ThemeColorController>();
         }
 
-        private void SetTransformSync(Transform transf, ControllerSyncMode syncMode, ControllerSyncMode targetMode)
+        private void UpdateSyncMode(Transform inputTransform, ControllerSyncMode userSyncMode, ControllerSyncMode desiredSyncMode)
         {
 
             #if UDONSHARP
 
-            transf.GetComponent<UdonBehaviour>().enabled = syncMode < targetMode;
+                inputTransform.GetComponent<UdonBehaviour>().enabled = (int)userSyncMode < (int)desiredSyncMode;
 
             #endif
 
         }
 
-        void SetControllerSyncMode(ControllerSyncMode syncMode)
+        public void SetControllerSyncMode(ControllerSyncMode syncMode)
         {
             
-            SetTransformSync(gainSlider.transform, syncMode, ControllerSyncMode.ExcludePowerAndGain);
-            SetTransformSync(fadeLengthSlider.transform, syncMode, ControllerSyncMode.None);
-            SetTransformSync(fadeExpFalloffSlider.transform, syncMode, ControllerSyncMode.None);
+            UpdateSyncMode(gainSlider.transform, syncMode, ControllerSyncMode.ExcludePowerAndGain);
+            UpdateSyncMode(fadeLengthSlider.transform, syncMode, ControllerSyncMode.None);
+            UpdateSyncMode(fadeExpFalloffSlider.transform, syncMode, ControllerSyncMode.None);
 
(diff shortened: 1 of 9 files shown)


Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
```diff
@@ -109,6 +110,41 @@ ThemeColorController FindThemeColorController()
...
+            #endif
+
+        }
+
+        void SetControllerSyncMode(int syncMode)
+        {
+            
+            SetTransformSync(gainSlider.transform, syncMode, 2);
+            //SetTransformSync(trebleSlider.transform, syncMode, 2);
+            //SetTransformSync(bassSlider.transform, syncMode, 2);
```
float3, line 129:
```suggestion

```

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
```diff
@@ -109,6 +110,41 @@ ThemeColorController FindThemeColorController()
...
+
+            #endif
+
+        }
+
+        void SetControllerSyncMode(int syncMode)
+        {
+            
+            SetTransformSync(gainSlider.transform, syncMode, 2);
+            //SetTransformSync(trebleSlider.transform, syncMode, 2);
```
float3, line 128:
```suggestion

```

float3 (approved):
this looks good to me but pema would be the person to decide if this is a feature that we want
have you had a conversation with him?

fundale:
I have not yet

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
```diff
@@ -109,6 +110,39 @@ ThemeColorController FindThemeColorController()
...
+            transf.GetComponent<UdonBehaviour>().enabled = syncMode < targetMode;
+
+            #endif
+
+        }
+
+        void SetControllerSyncMode(int syncMode)
+        {
+            
+            SetTransformSync(gainSlider.transform, syncMode, 2);
```
pema99, line 127: Why the magic numbers? Are these supposed to match `ControllerSyncMode`? If so, please use the enum type throughout. `<` etc are defined for enums

Packages/com.llealloo.audiolink/Runtime/Scripts/GlobalToggle.cs
```diff
@@ -29,13 +29,17 @@ private void Start()
 
         public override void OnDeserialization()
         {
+            if (!enabled) return;
```
pema99, line 32: I guess this is fine, but out of curiosity, which issue were hitting without this?

fundale: The `if !enabled` on Toggle and Slider Update is for handling a state where Unity still calls the function with the behavior disabled, also added to OnDeserialization just to be sure, never actually checked to see if it caused a difference

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
fundale: Yes, didn't see that before

float3 (approved):

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
```diff
@@ -109,6 +110,39 @@ ThemeColorController FindThemeColorController()
             return controllerTransform.GetComponent<ThemeColorController>();
         }
 
+        private void SetTransformSync(Transform transf, ControllerSyncMode syncMode, ControllerSyncMode targetMode)
+        {
+
+            #if UDONSHARP
+
+            transf.GetComponent<UdonBehaviour>().enabled = syncMode < targetMode;
```
float3, line 118: can you explain this?

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
```diff
@@ -109,6 +110,39 @@ ThemeColorController FindThemeColorController()
...
+
+            #endif
+
+        }
+
+        void SetControllerSyncMode(ControllerSyncMode syncMode)
```
pema99, line 124: This should probably be public so user scripts can call it.

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
pema99: The enum is an ordering. It lets you exclude _just_ the power and gain from being synced (seems useful to me). So for the gain slider, targetMode is ExcludePowerAndGain. Anything less than that (ie. ExcludePower and All) mean that the gain slider should be synced.

float3: i meant the GetComponent line specifically

fundale: The weird naming of the function and variables?
Idk, but I can try and come up with some more descriptive and accurate names.

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
```diff
@@ -109,6 +110,39 @@ ThemeColorController FindThemeColorController()
...
+
+            #if UDONSHARP
+
+                inputTransform.GetComponent<UdonBehaviour>().enabled = (int)userSyncMode < (int)desiredSyncMode;
+
+            #endif
+
+        }
+
+        public void SetControllerSyncMode(ControllerSyncMode syncMode)
```
pema99, line 124: This seems to be missing some elements, such as everything relating to theme color.

fundale: Included ThemeColorController in other toggles sync, lmk if there's something else I forgot to include

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
fundale: SetControllerSyncMode has been made public

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkController.cs
fundale: The UdonBehaviour generalization is to avoid having two functions, one for GlobalSlider and GlobalToggle,
However I can make it two separate functions if desired

pema99:
Just tested this locally. Seems to work nicely. Thanks for the contribution.

PR #313, opened 2024-05-18
Try and Catch on Unlink

fundale:
Attempts to recover from a invalid `UnlinkAnalyser()` call


float3:
do you wanna tackle color chord in the same PR?

fundale:
> do you wanna tackle color chord in the same PR?

Looked into it,
ColorChord appears to be functioning, but under the threshold for the panel display,
the entire CC strip and notes are sparse, however I'm not familiar with how CC works internally.
cnlohr added it correct? maybe they have an idea on where to check?

float3:
if it's functioning what's the problem?

fundale:
Ok so, it's functioning in that it is returning a value, but not functioning in that it's return value is close to 0

float3:
oh interesting

PR #315, opened 2024-06-18
Reduce network size for mini player and theme controller

Happyrobot33:
Hello! Just a simple PR to fix the bloat cost of arrays, specifically for the theme colors. Change is based on my calculator here: https://docs.google.com/spreadsheets/d/1WXrHxJQW_evivVpajzLnrZnmu0otWxVsBsfUZiQXLFg/edit?usp=sharing

The simple difference this makes is going from 136 bytes for the theme controller down to either 100 or 92 bytes depending on how the client decides to sort the variable names. Let me know if you have any questions! I can also simplify this to not use an array at all if you'd like, but this was the most drop in solution as it keeps the array accessing logic completely the same


Happyrobot33:
there was another optimization I wanted to make but it would've changed the access level of a variable I'm pretty sure, so I haven't committed that in here

float3:
what variable?

Happyrobot33:
> what variable? these things are negotiable

The ones I was looking at are _syncOwnerPlaying and _syncLocked in the mini player. They should be packed into a byte instead of being individual booleans. I could do this, although the one variable being nonserialized makes it more complicated than I think its worth now thinking about it

float3:
couldn't we keep them the way they currently are and just serialize and deserialize them that way when we send them over the wire?

Happyrobot33:
> couldn't we keep them the way they currently are and just serialize and deserialize them that way when we send them over the wire?

well yes that would be the fundamental change. I can look into it deeper, it honestly might still be fine as a drop in property setup

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorController.cs
```diff
@@ -16,13 +16,34 @@ public class ThemeColorController : MonoBehaviour
     {
         [UdonSynced] private int _themeColorMode;
 
+        public Color[] customThemeColors
+        {
+            get
+            {
+                return new[] {
```
pema99, line 23: somethings gone funky with the indentation here I think

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorController.cs
```diff
@@ -16,13 +16,34 @@ public class ThemeColorController : MonoBehaviour
     {
         [UdonSynced] private int _themeColorMode;
 
+        public Color[] customThemeColors
```
pema99, line 19: big unfortunate this is now a footgun property, like
```
controller.customThemeColors[0] = Color.white;
```
will no longer work since you write to a copy. But I guess if it is in the name of optimizing network traffic, it's justified

Happyrobot33: Sadly true, i hadnt thought about this, and that makes it a little weird to work with. Maybe just not using the array at all should be done to prevent confusion

pema99: Perhaps could mark it as deprecated and change our own code to not use it, if you can be bothered

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorController.cs
Happyrobot33: in what way? not sure I understand

Happyrobot33:
I have changed over to dedicated methods and obsoleted the array. I also went ahead and made the change to the mini player, but would like a more thorough testing of it as I'm not fully aware of the codebase and am unsure if I made the proper changes to fully support it.

Happyrobot33:
bump on review

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorControllerV0.cs
```diff
@@ -16,13 +17,27 @@ public class ThemeColorControllerV0 : MonoBehaviour
     {
         [UdonSynced] private int _themeColorMode;
 
+        [Obsolete("This array will return a copy of the data, causing it to not write any data when trying to write directly to a index. Use " + nameof(GetCustomThemeColors) + " and " + nameof(SetCustomThemeColors) + " instead.", true)]
```
float3, line 20: hold on why do we need this when you implemented getter and setter correctly?

float3 (commented):
wow this is some spaghetti code in the name of optimization, lgtm but we need to wait for pema to check and we should probably test it

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorControllerV0.cs
Happyrobot33: Because if someone was to directly index to set a value, it wouldnt trigger the setter.

So array[2] = false wouldnt trigger the setter for array, annoyingly

float3: got it thanks for the clarification

Happyrobot33:
> wow this is some spaghetti code in the name of optimization, lgtm but we need to wait for pema to check and we should probably test it

If you have any suggestions on what youd like to see cleaned up id be willing to try tp figure it out. Ive done some minimal testing but i dont use the features touched here extensively so yes more testing would be nice

pema99:
I'll take another look this weekend. Thanks for reminding

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorController.cs
```diff
@@ -170,17 +216,15 @@ public void InitializeAudioLinkThemeColors()
...
             customThemeColors[3] = audioLink.customThemeColor3;
 
-            _initCustomThemeColors = new[] {
-                customThemeColors[0],
-                customThemeColors[1],
-                customThemeColors[2],
-                customThemeColors[3],
-            };
+            //shallow copy of the array
+            _initCustomThemeColors = GetCustomThemeColors();
```
pema99, line 227: this isn't a shallow copy since Color is a value type?

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMiniPlayer.cs
```diff
@@ -50,15 +50,26 @@ public class AudioLinkMiniPlayer : UdonSharpBehaviour
...
+        /// <summary>
+        /// Synced via <see cref="_flags"/>
+        /// </summary>
         bool _syncLocked = true;
 
+        /// <summary>
+        /// Flags byte for various settings that are synced. Order is <see cref="_syncOwnerPlaying"/>, <see cref="_syncLocked"/>
+        /// </summary>
+        [UdonSynced]
+        byte _flags = 0;
```
pema99, line 71: Maybe make the default value for this match the default value for the fields? ie 0b10

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorController.cs
pema99: It's fine that you've deprecated this, but please don't make it an error (passing true to ObsoleteAttribute). That would make this a breaking change. Let's save those for later.

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorController.cs
```diff
@@ -106,11 +152,11 @@ public void OnGUIchange()
             if (!Networking.IsOwner(gameObject))
                 Networking.SetOwner(localPlayer, gameObject);
 #endif
-            customThemeColors[customColorIndex] = Color.HSVToRGB(
+            SetCustomThemeColors(GetCustomThemeColors()[customColorIndex] = Color.HSVToRGB(
```
pema99, line 155: Nitpick, but could you split this into a few statements? This is begging to be misread

pema99:
Sorry, got sidetracked again. Looks fine overall but I had some additional comments

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorController.cs
Happyrobot33: isnt this somewhat a breaking change though? if people are trying to directly set a index it wont work anymore which is breaking correct?

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorController.cs
Happyrobot33: right but the array isnt a reference to the real array here, its a copy of the array itself. I could just remove the word shallow

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMiniPlayer.cs
Happyrobot33: I changed it to that, however its not a big deal as it is also set in start() to ensure that it is correct

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorController.cs
pema99: Ah yeah, that's a fair point. It's behaviorally a breaking change. Since we only switched to semver a few releases ago, we haven't talked about if changes like that (where the API doesn't change but the behavior does) should be considered backwards compatible or not (ie bump major or minor version)

pema99: Regardless it shouldn't need to block this PR

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorController.cs
pema99: sounds good to me

Packages/com.llealloo.audiolink/Runtime/Scripts/ThemeColorController.cs
float3: wouldnt consider it backwards compatible imo

float3: we can mark the setter as deprecated maybe? how good is unities deprecation and updating system

Happyrobot33: From my understanding and research ive done in the past, the popup that can appear when you use built in deprecated code is somehow either hard coded or baked into something in the engine and there doesnt seem to be a attribute to have the same behaviour :(

PR #318, opened 2024-07-25
Add support for dual mono setups (AVPro)

fundale:
Optional support for dual mono audio

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
@@ -759,6 +759,12 @@ namespace AudioLink
             InitIDs();
             audioSource.GetOutputData(_audioFramesL, 0);                // left channel
 
+#if UDONSHARP
+            bool hasDualMono = VRC.SDKBase.Utilities.IsValid(optionalRightAudioSource);
+#else
+            bool hasDualMono = optionalRightAudioSource != null;
+#endif
+
             if (_rightChannelTestCounter > 0)
             {
                 if (_ignoreRightChannel)
@@ -767,7 +773,7 @@ namespace AudioLink
                 }
                 else
                 {
-                    if (VRC.SDKBase.Utilities.IsValid(optionalRightAudioSource))
+                    if (hasDualMono)
                     {
                         optionalRightAudioSource.GetOutputData(_audioFramesR, 0);
                     } else audioSource.GetOutputData(_audioFramesR, 1);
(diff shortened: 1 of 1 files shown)


Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -765,15 +767,21 @@ public void SendAudioOutputData()
                 }
                 else
                 {
-                    audioSource.GetOutputData(_audioFramesR, 1);
+                    if (VRC.SDKBase.Utilities.IsValid(optionalRightAudioSource))
```
float3, line 770: issue (blocking): any calls to VRC namespace stuff should be wrapped with `#if UDONSHARP` or `#if VRC_SDK_VRCSDK3` depending on if that codepath should be taken in Avatar Projects, in this case I believe it should be `UDONSHARP`

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -765,15 +767,21 @@ public void SendAudioOutputData()
...
                 _rightChannelTestCounter--;
             }
             else
             {
-                _rightChannelTestCounter = _rightChannelTestDelay;      // reset test countdown
-                _audioFramesR[0] = 0f;                                  // reset tested array element to zero just in case
-                audioSource.GetOutputData(_audioFramesR, 1);            // right channel test
+                _rightChannelTestCounter = _rightChannelTestDelay;                  // reset test countdown
+                _audioFramesR[0] = 0f;                                              // reset tested array element to zero just in case
+                if (VRC.SDKBase.Utilities.IsValid(optionalRightAudioSource))        // check if dual mono is present
```
float3, line 781: issue (blocking): any calls to VRC namespace stuff should be wrapped with `#if UDONSHARP` or `#if VRC_SDK_VRCSDK3` depending on if that codepath should be taken in Avatar Projects, in this case I believe it should be `UDONSHARP`

float3 (changes requested):

float3 (approved):
lgtm

pema99:
The change looks fine to me, but could you enlighten me a bit on how this should be used? It looks like it will just always pull the data from the newly added audio source field if it isn't null. Isn't this for switching between 2 audio sources? How do you actually switch?

fundale:
> @fundale The change looks fine to me, but could you enlighten me a bit on how this should be used? It looks like it will just always pull the data from the newly added audio source field if it isn't null. Isn't this for switching between 2 audio sources? How do you actually switch?

This isn't for switching between two separate videoplayers or audiosources,
it's intention is to provide a way for video players using separate left and right speakers to function better.
Like when using AVPro through ProTV, VideoTXL, USharpVideo, etc.

As some audio may be stereo encoded with temporary mono right audio, that currently would not be captured when setup with AVPro unless using Stereo Mix

pema99:
So how exactly is a user intended to use this? What do they plug into the optional right source slot? The only difference I see in code is the channel they sample from, and only one is used at any given time

fundale:
> So how exactly is a user intended to use this? What do they plug into the optional right source slot? The only difference I see in code is the channel they sample from, and only one is used at any given time

If they're using a 2D Stereo mix source nothing changes with how they use it,

But for users using AVPro with dual mono they would have the left audio source remain in the current Audio Source slot, with the right audio source going into the Optional Right Audio Source.

Some video players already automate connecting to AudioLink so they may be able to auto configure it accordingly for users

It should still sample from left and right at the same time and combine them the same way AudioLink already does

pema99:
Oh, I was misreading the diff. It looks fine to me with your explanation.

pema99 (approved):

PR #320, opened 2024-09-07
Add AudioReactiveBlendshapes.cs and related files

fundale:
A small utility script based off `AudioReactiveLight.cs` for blendshapes, allowing a user to quickly drive blendshapes by specifying their index in an array.

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveBlendshapes.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveBlendshapes.cs
@@ -6,8 +6,10 @@ namespace AudioLink
     using UdonSharp;
     using UnityEngine.Rendering.PostProcessing;
 
+    [RequireComponent(typeof(SkinnedMeshRenderer))]
     public class AudioReactiveBlendshapes : UdonSharpBehaviour
 #else
+    [RequireComponent(typeof(SkinnedMeshRenderer))]
     public class AudioReactiveBlendshapes : MonoBehaviour
 #endif
     {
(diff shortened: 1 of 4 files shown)


Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveBlendshapes.cs
```diff
@@ -0,0 +1,47 @@
...
+            _maxBlendshapes = Mathf.Min(maxBlendshapes);
+
+        }
+
+        void Update()
+        {
+            if (audioLink.AudioDataIsAvailable())
+            {
+                // Convert to grayscale
+                float amplitude = Vector3.Dot(audioLink.GetDataAtPixel(delay, band), new Vector3(0.299f, 0.587f, 0.114f)) * 100f;
```
pema99, line 40: Hmm this grayscale conversion looks sus to me. In the 4band section, each channel means a different thing (they are different levels of filtering, not chroma values). You should probably just select 1 single channel, and maybe put the channel select on a dropdown

pema99:
Thanks looks nice - check my comment

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveBlendshapes.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace AudioLink
+{
+#if UDONSHARP
+    using UdonSharp;
+    using UnityEngine.Rendering.PostProcessing;
+
+    public class AudioReactiveBlendshapes : UdonSharpBehaviour
+#else
+    public class AudioReactiveBlendshapes : MonoBehaviour
```
pema99, line 11: Use [RequireComponent(typeof(SkinnedMeshRenderer))] ?

fundale: SkinnedMeshRenderer is now required

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveBlendshapes.cs
fundale: Set it to access the pixel's red value for now,
I like the idea of having an option to select which channel to read from a pixel, though that may be better as a enum in a different file

PR #326, opened 2024-12-15
Minor GC improvement

techanon:
Hoist some arrays to class fields to avoid excessive array allocations (on behalf of @dooly123).

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
@@ -732,22 +732,22 @@ namespace AudioLink
         }
 #endif
 
-        private const int MaxLength = 32;
-        private int[] globalStringCodePoints = new int[MaxLength];
-        private const int packedVectorsLength = MaxLength / 4;
-        private Vector4[] globalStringPackedVectors = new Vector4[packedVectorsLength];
+        private const int GlobalStringMaxLength = 32;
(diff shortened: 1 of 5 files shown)


Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -732,17 +732,25 @@ public override void OnDeserialization()
         }
 #endif
 
+        private const int MaxLength = 32;
+        private int[] codePoints = new int[MaxLength];
+        public const int VecsLength = MaxLength / 4;
```
pema99, line 737:
```suggestion
        private const int VecsLength = MaxLength / 4;
```

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -732,17 +732,25 @@ public override void OnDeserialization()
...
+        private Vector4[] vecs = new Vector4[VecsLength];
```
pema99, line 738: Since this is now in class scope, maybe rename it to something more telling, like packedGlobalTextVecs

pema99: Along with the name of the length

pema99:
Looks good, check the comment and we can merge

float3:
if you edit one of the projects .gitignore please edit all of them

pema99:
Thank you (and thanks to dooly)

PR #324, opened 2024-10-26
Preliminary Unity 6 support

techanon:
This should fix some edge-case issues in the editor scripts when trying to pull the minimal build into a Unity 6 project.
I'm open to a more proper fix cause this does feel kinda goofy.

PR #328, opened 2024-12-22
Make use of FindObjectsByType consistent. 

techanon:
Anything beyond 2021.3 supports the ByType API.
Is additional cleanup from #324 changes.

### DIFF
--- a/Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkEditor.cs
+++ b/Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkEditor.cs
@@ -2,9 +2,8 @@
 using System.Reflection;
 
 #if UDONSHARP
-using UdonSharp;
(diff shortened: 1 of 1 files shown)


float3:
can you make it use the same style I did? I prefer that

float3:
otherwise lgtm

techanon:
Not sure what you mean by style, the api calls are completely different between the two.

float3:
if you give me access to the branch I'll fix it

techanon:
The PR has repo contributor changes enable. You should just be able to modify it.

techanon:
Your change is incorrect because the FindObjectsByType does not exist prior to 2021_3 ergo the whole line change there. Your style preference doesn't actually work for the older unity versions.

float3:
oh I aee, ByType vs OfType

PR #329, opened 2025-01-24
DFT phase

tt0fu:
Used the previously reserved dft alpha channel to store the the phase of the signal and added an example material that uses it.


Packages/com.llealloo.audiolink/Runtime/Shaders/AudioLink.shader
tt0fu, line 1: It's very strange that this giant list is here. If someone wants actual raw DFT data, they need to find this list in the CRT shader and copy it into their shader.

float3:
the diff looks a bit weird

tt0fu:
> the diff looks a bit weird

I apologize for the strange single space changes in the commits. My IDE added them automatically and I didn't notice them being included in the commits.

float3:
no worries, let's revert those

Packages/com.llealloo.audiolink/Runtime/Shaders/AudioLink.shader
float3: yeah I guess that was an oversight of the one who designed that API, thanks for fixing it

float3:
pema will have to look at and approve/reject this PR, I don't interact with the HLSL part of Audiolink much these days

Packages/com.llealloo.audiolink/Runtime/Shaders/AudioLink.shader
tt0fu: It's also very strange that the DFT algorithm itself walks through the samples backwards and specifically through 3069/2 (that's not even divisible by 2?) of them.

cnlohr: IIRC
1.  It was backwards since we want to heavily weight the newest samples.
2. How do you compute 3069/2?

cnlohr:
One other thought here.  The operation being done here can be completely captured by using goertzel's algorithm which is likely to be moderately faster in an application like this, especially on mobile hardware. I just haven't gotten around to testing it.

What are you using phase for?  (Humans can't really perceive phase) Just some additional visualization?  OK Either way on the getting-the-phase information.  I just never found a use.  I tried doing like sound reconstruction but, because the bins are spaced for human perception instead of DSP, you end up getting goop out when you try to reconstitute.

pema99 (approved):
code looks fine to me and charles seems happy with it, so merging

cnlohr:
:partying_face:

PR #333, opened 2025-05-16
Optimize audio sample array access for mobile GPU performance

Hamaguri-0414:
We found a problem with audio sample code that uses changing index numbers for array access. This prevents the compiler from optimizing the code well, causing the shader to fail on older Metal GPUs (before A14 Bionic).
This fix limits the array access range clearly, which helps the compiler make better optimizations.

We tested the fix and confirmed the shader now works properly on Metal GPUs older than A14 Bionic.

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Shaders/AudioLink.shader
+++ b/Packages/com.llealloo.audiolink/Runtime/Shaders/AudioLink.shader
@@ -221,29 +221,43 @@ Shader "AudioLink/Internal/AudioLink"
 
             float ReadLeft( int sample )
             {
-                if( sample < 1023 )
-                    return _Samples0L[sample];
-                else if( sample < 2046 )
-                    return _Samples1L[sample-1023];
-                else if( sample < 3069 )
-                    return _Samples2L[sample-2046];
-                else if( sample < 4092 )
-                    return _Samples3L[sample-3069];
-                else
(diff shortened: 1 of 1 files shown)


float3 (approved):

float3:
lgtm

pema99:
This really doesn't sit well with me. It's a de-optimization in almost all cases. Integer division and modulo are some of the absolute slowest operations you can do on a modern GPU. On top of that, the change increase the instruction count by around 7% and the register count by 50%. The code as it is now is pretty objectively better to me.

You write
> This prevents the compiler from optimizing the code well, causing the shader to fail on older Metal GPUs (before A14 Bionic).

It's unclear to me not "optimizing the code well" translates into the shader failing. Do you actually get a compilation error? If so, which one? What if you slap a [branch] attribute on each of the if-statements? And what about with _just_ adding the clamp?

Also p.s, your bound check is wrong. When passing index 4092, the check `4092 < sample` will return false, and you will index out of bounds. Should be `<=`.

pema99 (changes requested):
I'd like to understand the issue you are addressing better before we consider merging this. If we do end up merging it, it should probably be behind an #ifdef for Metal only. Check my previous comment.

pema99:
Marking as draft as this hasn't had any activity for quite a while.

PR #337, opened 2025-06-03
Some utility script cleanups

fundale:
Custom Editor GUI for AudioReactive scripts,
Disable network sync on AudioLinkZone,
And a few more API functions.


Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkReactiveEditor.cs
```diff
@@ -0,0 +1,261 @@
...
+                        Mesh blendshapeMesh = blendshapeMeshRenderer.sharedMesh;
+
+                        if (blendshapeMesh.blendShapeCount != blendshapeNames.Count)
+                        {
+                            blendshapeNames.Clear();
+                            for (int indx = 0; indx < blendshapeMesh.blendShapeCount; indx++)
+                                blendshapeNames.Add(blendshapeMesh.GetBlendShapeName(indx));
+                        }
+
+                        SerializedProperty blendshapeIDs = serializedObject.FindProperty("blendshapeIDs");
```
pema99, line 128: can we use `nameof()` in cases like this?

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkReactiveEditor.cs
```diff
@@ -0,0 +1,261 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using System;
+using UnityEditorInternal;
+
+namespace AudioLink.Reactive
+{
+
+    [CustomEditor(typeof(AudioReactive), true)]
+    public class AudioLinkReactiveEditor : UnityEditor.Editor
```
pema99, line 12: I appreciate the effort you've put into this, it looks nice, but I don't really get why you've stuffed all the editor code into one big file. It's a bit difficult to read, and also leads to a bunch of special-casing. Do you think we could refactor this a bit to just have some helper functions that are agnostic of the specific type being edited, and then have a separate editor for each type? 

If not, I'm not totally against this, especially considering it is just UI code. Lmk your thoughts

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.DataAPI.cs
```diff
@@ -185,6 +185,26 @@ public Vector4 GetDataAtPixel(Vector2 position)
...
+        /// <param name="delay">How long to wait (0 - 127), or if smoothed how smooth (0 - 15).</param>
+        /// <param name="smooth">Wheather to smooth the output.</param>
+        /// <returns>Float value from History or Smoothed.</returns>
+        public float GetBandAsSmooth(AudioLinkBand band, int delay, bool smooth)
+        {
+            int bandID = (int)band;
+
+            if (smooth)
+            {
+                bandID += 28;
```
pema99, line 201: Use  GetALPassFilteredAudioLink()?

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveSurface.cs
```diff
@@ -65,13 +64,21 @@ void Start()
...
             MaterialPropertyBlock block = new MaterialPropertyBlock();
-            MeshRenderer mesh = GetComponent<MeshRenderer>();
-            block.SetFloat(_Delay, (float)delay / 128f);
-            block.SetFloat(_Band, (float)band);
+            MeshRenderer mesh = transform.GetComponent<MeshRenderer>();
+            int bandInt = (int)band;
+            int delayTmp = delay;
+            if (smooth)
+            {
+                bandInt += 28;
```
pema99, line 72: same thing here, we have a constant in the dataapi for this

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveSurfaceArray.cs
```diff
@@ -78,6 +77,14 @@ void Start()
         public void UpdateChildren()
         {
 
+            MaterialPropertyBlock block = new MaterialPropertyBlock();
```
pema99, line 80: did you test that this actually works when the property block is modified for every renderer? I'm guessing renderer.SetPropertyBlock() makes a copy? If so, cool :)

pema99:
I left some comments, have a look

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveSurfaceArray.cs
fundale: Appears to,
Delay step is set incrementally for each child.
image

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveBlendshapes.cs
```diff
@@ -1,32 +1,31 @@
 ﻿using UnityEngine;
 
-namespace AudioLink
+namespace AudioLink.Reactive
```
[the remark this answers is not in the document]
fundale: I'll revert the Namespace

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.DataAPI.cs
```diff
@@ -185,6 +185,26 @@ public Vector4 GetDataAtPixel(Vector2 position)
...
+                delay = 15 - delay;
```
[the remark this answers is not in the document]
fundale: The Filtered AudioLink output is 16 pixels wide, with smoothest on the *Left* side (x = 0) and most responsive on the *Right* side (x = 15),
this inverts the delay value when it's being used as a smoothness value to make it "smoother" the higher the value.

*(There's no value that provides the dimensions of passes currently in the DataAPI.)*

Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkReactiveEditor.cs
fundale: I was thinking of having separate Inspectors for each originally,
But didn't want to clutter up the Editors folder, thinking now I could make a subfolder and a helper class...

fundale: I'll look into this...

fundale:
lmk if the Split Editors are satisfactory or need more adjustments, then I can remove the Unified Editor

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.DataAPI.cs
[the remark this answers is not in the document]
pema99: In that case, we should probably add a new entry to the DataAPI :)

pema99:
@fundale The split editors look like an improvement to me. There's still some special casing in the common code, matching on a field called "delay" and "smooth". Ideally we'd get rid of that too, but if you find it too impractical to factor out, don't bother. This is still a lot nicer, I think.

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.DataAPI.cs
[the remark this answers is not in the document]
fundale: Alright, would it be worth adding these to the .cginc too?
If so maybe a later commit.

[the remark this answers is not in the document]
pema99: Probably but you can also leave it, either is fine

pema99:
@fundale Ok, thanks for trying. I think this is more or less good to go, though some merge conflicts have popped up (sorry!). Fix them and we should be good to merge.

PR #341, opened 2025-06-26
Add audiosource curve support to volume assignment (both falloff and blend)

techanon:
Audiolink normalizes the DFT by dividing by volume. This includes adjustment to the volume based on the falloff/blend curves.
In standalone, it will search for a valid (active and enabled) AudioListener and assume that is the audioTarget to calculate distance curve value from.
In VRC, it uses the player's head tracking data.

Additionally, some logic was added to make sure that playmode doesn't permanently modify the update mode from what it is set to originally. This is necessary for when a separate CRT wraps the original audiolink CRT (such as what VRSL does).

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
@@ -1,5 +1,5 @@
 ﻿using System;
-
+using System.Linq;
 using UnityEngine;
 
 namespace AudioLink
@@ -139,22 +139,29 @@ namespace AudioLink
 #if UDONSHARP
         private bool _hasInitializedTime = false;
         private VRCPlayerApi _localPlayer;
-#endif
-        private double _fpsTime = 0;
-        private int _fpsCount = 0;
-        private int _lastUpdatedFrame = -1;
-
-#if UDONSHARP
         private double GetElapsedSecondsSince2019() { return (Networking.GetNetworkDateTime() - new DateTime(2020, 1, 1)).TotalSeconds; }
         //private double GetElapsedSecondsSinceMidnightUTC() { return (Networking.GetNetworkDateTime() - DateTime.UtcNow.Date ).TotalSeconds; }
 #else
         private double GetElapsedSecondsSince2019() { return (DateTime.UtcNow - new DateTime(2020, 1, 1)).TotalSeconds; }
 #endif
 
+        private double _fpsTime = 0;
+        private int _fpsCount = 0;
+        private int _lastUpdatedFrame = -1;
+
         // Fix for AVPro mono game output bug (if running the game with a mono output source like a headset)
         private int _rightChannelTestDelay = 300;
         private int _rightChannelTestCounter;
         private bool _ignoreRightChannel = false;
+        private CustomRenderTextureUpdateMode initialUpdateMode = CustomRenderTextureUpdateMode.Realtime;
+
+#if UDONSHARP
+        [HideInInspector, SerializeField] private Transform audioTarget;
+        [HideInInspector, SerializeField] private bool autoDetectAudioTarget = false;
+#else
+        public Transform audioTarget;
+        public bool autoDetectAudioTarget = true;
+#endif
 
         #region PropertyIDs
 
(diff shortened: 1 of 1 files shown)


Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -370,13 +376,23 @@ void Start()
...
+            initialUpdateMode = audioRenderTexture.updateMode;
 
             // Disable camera on start if user didn't ask for it
             if (!audioDataToggle)
             {
                 DisableReadback();
             }
+
+#if !UDONSHARP && !CVR_CCK_EXISTS
+            Invoke(nameof(AutoCacheAudioTarget), 1);
```
pema99, line 388: Why the Invoke delay?

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -370,13 +376,23 @@ void Start()
...
+#endif
+        }
+
+        void OnDestroy()
+        {
+            // makes sure that playmode doesn't permanently modify the update mode
+            audioRenderTexture.updateMode = initialUpdateMode;
```
pema99, line 395: I wonder if you can just EditorUtility.ClearDirty(audioRenderTexture) after modifying the update mode 🤔

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -545,10 +561,21 @@ private void Update()
...
-                audioMaterial.SetVector(_SourcePosition, audioSource.transform.position);
+                float sourceVolume = audioSource.volume;
+                Vector3 sourcePosition = audioSource.transform.position;
+                float sourceMax = audioSource.maxDistance;
 
+                AnimationCurve sourceBlendCurve = audioSource.GetCustomCurve(AudioSourceCurveType.SpatialBlend);
+                AnimationCurve sourceFalloffCurve = audioSource.GetCustomCurve(AudioSourceCurveType.CustomRolloff);
+                var playerEars = GetHearingLocation();
+                float listeningDistance = Vector3.Distance(playerEars, sourcePosition) / sourceMax;
+                float sourceFalloff = sourceFalloffCurve.Evaluate(listeningDistance);
```
pema99, line 572: Does this just clamp when the input is greater than 1? Can happen when past the max distance, right?

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -545,10 +561,21 @@ private void Update()
...
+                float sourceBlend = sourceBlendCurve.Evaluate(listeningDistance);
+                sourceVolume = Mathf.Lerp(sourceVolume, sourceVolume * sourceFalloff, sourceBlend);
```
pema99, line 574: I'm not too familiar with this API, so just curious: Are you mimicking behavior that is specified somewhere? Since we were just using audioSource.volume before. Or is this empirical?

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -611,6 +638,47 @@ private void Update()
...
+        }
+
+#if !UDONSHARP && !CVR_CCK_EXISTS
+        private double _lastAudioTargetCheckTime;
+
+        private void AutoCacheAudioTarget()
+        {
+            if (!autoDetectAudioTarget) return;
+            CacheAudioTarget();
+            Invoke(nameof(AutoCacheAudioTarget), audioTarget ? 10 : 1); // check faster until one is found
```
pema99, line 661: This I am really struggling to understand :P

pema99:
Change looks good overall, and I like the approach taken. I left some comments.. mostly questions, I'm a bit confused about some of the specifics, tbh. Check them and I think we can merge

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
techanon: Related to `AutoCacheAudioTarget`. See https://github.com/llealloo/audiolink/pull/341/files#r2193355775

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
techanon: It does not appear to be the case.

PR #342, opened 2025-06-26
Add theme color support to AudioReactiveLight.

techanon:
Also add script file for storing useful audiolink enums.

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkEnums.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkEnums.cs
@@ -1,6 +1,3 @@
-// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-
 using UnityEngine;
 
 namespace AudioLink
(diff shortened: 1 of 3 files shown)


Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveLight.cs
```diff
@@ -1,19 +1,23 @@
...
+using UdonSharp;
+#endif
+
 namespace AudioLink
 {
+    [RequireComponent(typeof(Light))]
 #if UDONSHARP
-    using UdonSharp;
-
+    [UdonBehaviourSyncMode(BehaviourSyncMode.NoVariableSync)]
```
pema99, line 11: Is this intentional?

pema99:
Change looks good, but what's up with that license text?

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioReactiveLight.cs
techanon: Yes. The behaviour has no sync data on it, so there is no need to have the inspector display the sync data mode options. NoVariableSync allows other scripts on the object to use sync data without interferring, unlike `None` mode.

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkEnums.cs
```diff
@@ -0,0 +1,24 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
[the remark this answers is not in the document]
techanon: huh. honestly no idea. removed.

techanon:
Actually, figured it out. Rider is generating that comment at the top of files that are created via Rider's add script menu. I updated rider recently so I guess it started doing that.

pema99 (approved):

PR #343, opened 2025-07-01
Implement Media Playback Reporting API in Bundled Mini Player and YTDLP player

JLChnToZ:
As Media Playback API was already exists for a year, I'm wondering why bundled video players didn't implement it.

This PR attempts to fills this gap.

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMiniPlayer.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMiniPlayer.cs
@@ -32,11 +32,14 @@ namespace AudioLink
         [Tooltip("Automatically loop track when finished")]
         public bool loop = false;
 
+        public AudioLink audioLink;
+
         [Header("Internal")]
         [Tooltip("Use this texture as an input to materials and other shader systems like LTCGI.")]
         public CustomRenderTexture videoRenderTexture;
         [Tooltip("AVPro video player component")]
         public VRCAVProVideoPlayer avProVideo;
+        public AudioSource mainAudioSource;
 
         float retryTimeout = 6;
         float syncFrequency = 5;
@@ -123,6 +126,8 @@ namespace AudioLink
 
                 _PlayVideo(defaultUrl);
             }
+
+            SyncPlayerStateToAudioLink();
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMiniPlayer.cs
```diff
@@ -32,11 +32,14 @@ public class AudioLinkMiniPlayer : UdonSharpBehaviour
         [Tooltip("Automatically loop track when finished")]
         public bool loop = false;
 
+        public AudioLink audioLink;
```
pema99, line 35: Did you check if this field works with the button on the AudioLink prefab to automatically populate references to it? Would be nice if you don't have to populate this manually.

Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
```diff
@@ -59,11 +60,27 @@ public enum Resolution
 
         void OnEnable()
         {
+            if (audioLink == null)
+            {
+                audioLink = GetComponentInParent<AudioLink>();
+                if (audioLink == null)
+                {
+                    audioLink = FindObjectOfType<AudioLink>();
```
pema99, line 68: It looks like this is deprecated since 22.3 (https://docs.unity3d.com/6000.1/Documentation/ScriptReference/Object.FindObjectOfType.html) and one should use Object.FindFirstObjectByType ?

pema99 (approved):
Thanks for this, change looks great overall! Check those few comments I left, and I can merge it.

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLinkMiniPlayer.cs
JLChnToZ: Yes it works (just checked in sandbox project)

Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
JLChnToZ: Oh haven't realized they have something new. Yes it can be replaced with `FindFirstObjectByType`. Should I change it or you will change it while merging?

pema99: I can do it

PR #350, opened 2025-12-22
Automate adjustments of Amplify Shaders

### LINKED ISSUE #348: Amplify Shaders are invisible in URP
**Describe the bug in detail:**
Amplify Shaders render invisible in URP.

**Provide steps/code to reproduce the bug:**
Enter play mode in a URP project.

**Expected behavior:**
The Amplify Shaders should render normally.

**Additional Information:**
I believe this is due to missing a "DepthOnly" pre-pass,
as I had different depth priming settings configured during development.

fundale:
Switches the "LightMode" tags depending on BiRP or URP, HDRP? pipelines.

Can switch between BiRP and URP with the same folder and readjust, with the caveat of trace white spaces, which I have not committed here.
Wouldn't call it Git "friendly", due to the trace white spaces.

Fixes #348
Reverts commit [bc1d4f3e0fbf19079c0de5f1285220057784fe3d](https://github.com/llealloo/audiolink/pull/335/commits/bc1d4f3e0fbf19079c0de5f1285220057784fe3d)

### DIFF
--- a/Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
+++ b/Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
@@ -3,6 +3,7 @@ using System.IO;
 using System.Text.RegularExpressions;
 using UnityEditor;
 using UnityEngine;
+using UnityEngine.Rendering;
 
 namespace AudioLink.Editor
 {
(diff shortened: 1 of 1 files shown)


Packages/com.llealloo.audiolink/Editor/Scripts/AudioLinkShaderCompatabilityUtility.cs
```diff
@@ -121,5 +123,85 @@ private static void ReplaceInFile(string path)
...
+
+        private static string SetLightModeTag(string text, string from, string to)
+        {
+            return new Regex($"\"LightMode\"\\s*=\\s*\"{from}\"").Replace(text, $"\"LightMode\"=\"{to}\"");
+        }
+
+        [MenuItem("Tools/AudioLink/Convert Amplify Shaders Pipeline")]
+        public static void ConvertShaderFiles()
+        {
+            ShaderInfo[] shaders = ShaderUtil.GetAllShaderInfo();
```
pema99: We might want to pop up an infobox when running this that informs the user that we are about to make changes to files on disk. It's a bit intrusive. Packages are supposed to be immutable so this is a bit strange, tbh, but I understand that there isn't really another way.

PR #351, opened 2025-12-29
Account for Color Theme mode 2

fundale:
Fixes the controller's behavior on the new "persistent Color Chord" mode 2 for ColorTheme


pema99:
Ah good catch, thanks

PR #354, opened 2026-01-31
AudioReactive PostProcess Script

fundale:
AudioReactive script that drives the Weight of a `PostProcessingVolume` determined by a set range


fundale:
> the asmdef change adds Unity.Postprocessing.Runtime as a hard reference. can you check what a project without com.unity.postprocessing installed does with that?

Missed an \#ifdef in the Editor, Otherwise seems to behave in a project without Post Processing,

Installed Post Processing and the Script appeared after a Domain reload.
After removing Post Processing, Script Define and another Domain reload the Script is removed again.

Pondering making a shell class for when it's disabled, but the Post Processing reference would be broken anyways...

Strangely no Errors, or even Warnings?

fundale:
Also may wanna run this by @pema99 And make sure it doesn't break like the previous attempt

Issue #357, opened 2026-02-12
Custom YTDL Location no longer works in 3.1.x


BluWizard10:
**Describe the bug in detail:**
For some reason, having a Custom YTDL Location set no longer works, now reverting back to the YTDL provided by VRChat by default at all times.

**Provide steps/code to reproduce the bug:**
Use AudioLink v3.1.0 or newer and attempt to use Custom YTDL Location. In my case, it was already configured in a previous version.

**Expected behavior:**
The Custom YTDL Location should be used and should show in the Logs.

pema99:
Hmm, this probably happened in the refactor of the ytdlpPlayer to support ffmpeg transcoding

PR #358, opened 2026-02-12
HOTFIX: Restore Custom YTDL Location and Custom FFmpeg Location Functionality.

BluWizard10:
### This is a critical fix!

Make sure that `FetchEditorPrefs();` is run when locating Ytdlp and FFmpeg locations. It is very important to make sure this is run. Otherwise, `_cachedEditorPrefs` is completely nullified/empty and the custom location is ignored.

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/ytdlpPlayer.cs
@@ -404,6 +404,9 @@ namespace AudioLink
         {
             _ytdlpFound = false;
 
+            // CRITICAL: Ensure cached prefs are up-to-date. Otherwise, custom path may be ignored after reload.
+            FetchEditorPrefs();
+
             // check for a custom install location
             string customPath = _cachedEditorPrefs.ytdlpPath;
             if (!string.IsNullOrEmpty(customPath))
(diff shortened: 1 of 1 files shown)


BluWizard10:
If merged, this will automatically Fix #357

fundale:
Seems to function properly,
Will throw an (benign) error when either `Locateytdlp()` or `LocateFFmpeg()` are called without the yt-dlp Player selected in the inspector, yet still functions as it pulls from `_cachedEditorPrefs` from previous `FetchEditorPrefs()` calls...

BluWizard10:
> Seems to function properly, Will throw an (benign) error when either `Locateytdlp()` or `LocateFFmpeg()` are called without the yt-dlp Player selected in the inspector, yet still functions as it pulls from `_cachedEditorPrefs` from previous `FetchEditorPrefs()` calls...

I seemed to have found a way to suppress that by catching it from the main thread. Added 2 commits.

fundale:
Yep, that cleaned up the Error, and it still functions fine

pema99:
I appreciate the extra effort with updating the version numbers, but please don't do this in the future. Fix looks good, thanks

BluWizard10:
> I appreciate the extra effort with updating the version numbers, but please don't do this in the future. Fix looks good, thanks

Noted, thanks.

PR #360, opened 2026-02-15
Add public wrapper property AudioLinkEnabled

CompuGeniusCode:
This allows us to set and access AudioLink's enabled state from our own scripts.


pema99:
fine addition

PR #362, opened 2026-02-21
Document frame counter

yum-food:
Audiolink's frame counter is useful for certain shaders, s.a. quasi-monte carlo fog simulations. It would benefit community shader developers for this to be documented.

I marked the other three fields as "Internal" (similar to the "Internal Timing Tracking" used in the same table) because the semantics are too nuanced to easily capture.


pema99 (approved):

pema99:
nice

PR #364, opened 2026-04-18
PoligonVR Support

TekkyNeko:
Adds support for the platform PoligonVR. I had ported this back in 2024, however it was in a destructive way that couldn't be contributed. So I decided to rewrite the port to make it properly friendly with the live build!

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
@@ -3,6 +3,12 @@ using UnityEngine;
 
 namespace AudioLink
 {
+#if !UDONSHARP
+	using Unity.Collections;
+	using UnityEngine.Rendering;
+	using static Shader;
+#endif
+
 #if UDONSHARP
     using UdonSharp;
     using VRC.SDK3.Rendering;
@@ -11,15 +17,18 @@ namespace AudioLink
 
     [UdonBehaviourSyncMode(BehaviourSyncMode.Manual)]
     public partial class AudioLink : UdonSharpBehaviour
-#else
-    using Unity.Collections;
-    using UnityEngine.Rendering;
-    using static Shader;
+#elif PVR_CCK_WORLDS
+	using PVR.CCK.Worlds.PSharp;
+	using System.Linq;
+
+	public partial class AudioLink : PSharpBehaviour
+#endif
 
 #if UNITY_WEBGL
     using System.Runtime.InteropServices;
 #endif
 
+#if !PVR_CCK_WORLDS && !UDONSHARP
     public partial class AudioLink : MonoBehaviour
 #endif
     {
@@ -105,8 +114,8 @@ namespace AudioLink
         public Color customThemeColor3 = new Color(0.0f, 1.0f, 0.0f, 1.0f);
 
         [Header("Custom Global Strings")]
-        [UdonSynced] public string customString1;
-        [UdonSynced] public string customString2;
+        [UdonSynced, PSharpSynced] public string customString1;
+        [UdonSynced, PSharpSynced] public string customString2;
 
         [HideInInspector] public Material audioMaterial;
         [HideInInspector] public CustomRenderTexture audioRenderTexture;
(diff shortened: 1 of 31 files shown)


TekkyNeko:
i just realized, I should probably clarify that ThemeColorController (and V0) dont correctly get setup on the stable build of PVR due to a bug with the SyncType, its cause of how i set it up, ([UdonSynced, PSharpSynced]), this is fixed on PVR's end for its next update.

float3:
Thanks for the PR

personay I'm for wider platform support and definitely want to merge your PR but we have no one on the team doing testing for anything other than VRC and Standalone, CVR and the web build targets are not tested and PVR would probably also not be tested by us so we would rely on you and the PVR community on Feedback, bug reports and fixes.

It seems there's some changes you accidentally committed, to keep the history clean please revert unnecessary changes like those to audiolink.mat

Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
```diff
@@ -105,8 +113,8 @@ public partial class AudioLink : MonoBehaviour
         public Color customThemeColor3 = new Color(0.0f, 1.0f, 0.0f, 1.0f);
 
         [Header("Custom Global Strings")]
-        [UdonSynced] public string customString1;
-        [UdonSynced] public string customString2;
+        [UdonSynced, PSharpSynced] public string customString1;
+        [UdonSynced, PSharpSynced] public string customString2;
```
float3: Do these need to be behind an if-guard?

TekkyNeko:
I went ahead and reversed the materials, those were accidental, likely partially cause i was testing in 2022.3.62f3. I don't mind testing and fixing the PVR support, as I've been doing that off and on for a while now already.

TekkyNeko:
All the random spaces to tabs should be fixed up now. Should be clean now?

pema99:
Hey 👋 

Thanks for your interest in the project and your effort on this. I'm honestly a bit conflicted with this PR. While it's always cool to see AudioLink running in more places, this is fairly large change adding several new bespoke behaviors just for PoligonVR. 

This is code which would, as float3 mentioned, not be tested by any of the primary maintainers. Such code not only has a tendency to rot, but also increases the cognitive load of making future changes.

For these reasons, I'm not particularly interested in maintaining support for every VR platform out there. Poligon is still very young, in early access, and has barely any users (no offense, I hope the platform succeeds!). To be frank, I don't the value in having this code in mainline AudioLink is very high. If someone had proposed adding support for CVR today, I think I would have said something similar - it was added during a time where AudioLink was in less of a "maintenance mode" than it is now (FWIW, the CVR integration also started its life as a separate fork).

May I suggest maintaining a fork for now? We'd gladly promote it in the README and on our discord. AudioLink updates and core feature additions are generally infrequent now, so maintaining a fork should be pretty manageable.

TekkyNeko:
sure, i can keep it in a separate fork like how it currently is/has been for the past 1.5 years. And ill try to keep it up with any updates. Thanks for the response!

pema99:
Ah, you mentioned it currently being a fork earlier, missed that :sweat_smile: 

I do think we should at least mention this in the Readme and Discord. Is this the correct one to promote? https://github.com/TekkyNeko/pvr-audiolink

TekkyNeko:
yep! thats the repo

float3:
the value might not be high but i think the cost is also very low since it's very easy to remove if we ever wish to remove it and it's easy to keep overview since everything added by this PR will have the `#if PVR_CCK_WORLDS` if guard

float3:
it adds 3 new files and less than 400 lines of code, i think it's fine to merge @pema99 but since you're the most active maintainer I'll leave the decision to you

pema99:
I think I stand by my reasoning. Sure, it's not an insurmountably large change, but it adds a cognitive burden to making future changes that isn't insignificant, fx:
- The #ifdef spam spread over the codebase already makes AudioLink tricky to modify and refactor. The preprocessor logic is now even more complex.
- Every time we want to add a synced variable, we have to add 2 separate attributes to it.
- The list paths in the include manager is yet another thing to keep in mind when adding, removing or renaming files 
- Several sections of the code are entirely duplicated, but with different types. This will easily go out of sync when changing those sections. It's also just confusing to see the same code twice.

We could fully embrace an approach of "ignore this code entirely during maintenance, and just let it break", but I'm not very comfortable with that. As a separate fork, the Poligon specific parts can be updated to accommodate changes to the codebase at its own pace.

pema99:
I've added mentions to the fork in the README for now. I'll mention it in the announcements channel one of the coming days, as well (assuming @TekkyNeko agrees).

As Poligon grows in the future, maybe I'll reconsider this. More alternatives to VRChat are always a great thing :)

TekkyNeko:
I'm fine with you announcing it in the discord. Again thanks for the response! 👍

PR #366, opened 2026-05-02
fix: re-publish _AudioTexture after render mode change

### LINKED ISSUE #365: `_AudioTexture` global stays null after a graphics-device reset
**Describe the bug in detail:**

AudioLink publishes `_AudioTexture` via `Shader.SetGlobalTexture` once at `OnEnable` and never republishes. Any event that resets Unity's per-graphics-device global state — most notably an XR loader cycle, but also other render-pipeline reconfigurations — wipes the binding. From that point on, every consumer that resolves the texture by global lookup gets `null`, with no recovery path in player builds.

**Provide steps/code to reproduce the bug:**

Repro environment: Unity 6000.4.x + URP 17.4.x + Basis Framework + OpenVR/SteamVR (Oculus Quest Pro over Virtual Desktop).

1. Add a normal AudioLink + AudioLinkController setup so `_AudioTexture` is published in Desktop mode.
2. Add a `MonoBehaviour` that logs `Shader.GetGlobalTexture("_AudioTexture")?.GetType().FullName` each `LateUpdate`.
3. Enter play mode. Log shows `UnityEngine.RenderTexture` every frame.
4. Hit F11 to trigger Basis's render-mode switch from Desktop to OpenVR _(the input action that calls `BasisDeviceManagement.SwitchSetMode`). Internally this calls `BasisXRManagement.StopXR` (`DeinitializeLoader`) and then `BasisXRManagement.LoadXR` (`InitializeLoader` + `StartSubsystems`), which causes Unity to reinitialise the graphics device for the new XR target._
5. From the first frame after `[SteamVR] Initialized` onward, the log shows `null`.

The same shape of bug should reproduce in any host that cycles XR loaders at runtime, not just Basis.

**Expected behavior:**

`_AudioTexture` should remain a valid `RenderTexture` for downstream consumers as long as AudioLink is enabled, even across graphics-device resets caused by XR loader cycles or other render-pipeline reconfigurations.

**Additional Information:**

**Suggested fix:**

*Layer 1.* Drop the `#if UNITY_EDITOR` gate on the `CacheAudioTarget` re-check so player builds also benefit from the periodic recovery. One-line change.

*Layer 2.* Add a `LateUpdate` (or `RenderPipelineManager.beginContextRendering`) republish while `_audioLinkEnabled` is true:

Either covers any host that resets globals.

**Basis Framework integration (optional, complementary):**

Basis already exposes `BasisDeviceManagement.OnBootModeChanged`, which fires after a render-mode switch completes. Subscribing to it in `OnEnable` and republishing in the handler gives a precise re-init signal without per-frame cost. I've prepared a branch demonstrating this on a fork:

<https://github.com/towneh/AudioLink/tree/feat-basis-xr-loader>

The Basis-specific code is gated behind `BASIS_FRAMEWORK_EXISTS`, so non-Basis projects are unaffected. 

Happy to get a PR going for the proven working Basis Framework approach in the meantime.

towneh:
### Summary

Subscribes AudioLink to `BasisDeviceManagement.OnBootModeChanged` (which already exists in Basis Framework) so `_AudioTexture` gets re-published after Basis switches render modes (Desktop ↔ OpenVR/OpenXR). Without this, the global binding gets wiped by the XR loader cycle and stays `null` for downstream consumers (custom URP passes, shaders that resolve the texture by global lookup) until something runs `SetAudioLinkGlobalTexture` again.

Closes #365

### What changes

- `Runtime/Scripts/AudioLink.cs` — subscribes to `BasisDeviceManagement.OnBootModeChanged` in `OnEnable`, unsubscribes in `OnDisable`, handler calls `SetAudioLinkGlobalTexture` while `_audioLinkEnabled` is true.
- `Runtime/AudioLink.asmdef` — adds `Basis Framework` to references so the Basis types are visible to the compiler when Basis is present.

All Basis-specific code is gated behind `BASIS_FRAMEWORK_EXISTS`, so projects without Basis compile this code out entirely. The asmdef reference to `Basis Framework` produces a benign warning in non-Basis projects (Unity tolerates missing asmdef references with "Reference to non-existent assembly" warnings; it doesn't fail compilation).

### Scope

This PR addresses only the Basis-specific integration. It does **not** include:

- Dropping the `#if UNITY_EDITOR` gate on the existing `CacheAudioTarget` re-check (lines 685–690).
- A host-agnostic per-frame republish via `LateUpdate` or `RenderPipelineManager.beginContextRendering`.

Either or both would complement this PR for hosts that don't expose a render-mode-changed event.

### Testing

Tested in a Unity 6000.4.3f1 + URP 17 + Basis Framework + Oculus Quest Pro (OpenVR/SteamVR) setup:

- **Before**: switching from Desktop to OpenVR wipes the global binding. Downstream consumers (a custom URP realtime light pass) freeze on initial state because `Shader.GetGlobalTexture("_AudioTexture")` returns `null` from the first frame after the switch onward.
- **After**: switching modes re-publishes the global. Downstream consumers continue functioning normally with audio reactivity intact.

Not yet tested in player builds (Editor playmode only), and not tested in non-Basis VR projects — the change is gated behind `BASIS_FRAMEWORK_EXISTS`, so non-Basis projects compile to current `master` behaviour.

### Notes

`BASIS_FRAMEWORK_EXISTS` is provided by Basis Framework itself; projects with Basis installed get the define automatically, and projects without it stay on the unchanged code path.

### DIFF
--- a/Packages/com.llealloo.audiolink/Runtime/AudioLink.asmdef
+++ b/Packages/com.llealloo.audiolink/Runtime/AudioLink.asmdef
@@ -4,7 +4,8 @@
     "references": [
         "UdonSharp.Runtime",
         "UdonSharp.Lib",
-        "VRC.Udon"
+        "VRC.Udon",
+        "Basis Framework"
     ],
     "includePlatforms": [],
     "excludePlatforms": [],
--- a/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
+++ b/Packages/com.llealloo.audiolink/Runtime/Scripts/AudioLink.cs
@@ -717,12 +717,27 @@ namespace AudioLink
         private void OnEnable()
         {
             EnableAudioLink();
+#if BASIS_FRAMEWORK_EXISTS
+            Basis.Scripts.Device_Management.BasisDeviceManagement.OnBootModeChanged
+                += OnBasisBootModeChanged;
+#endif
         }
 
         private void OnDisable()
         {
+#if BASIS_FRAMEWORK_EXISTS
+            Basis.Scripts.Device_Management.BasisDeviceManagement.OnBootModeChanged
+                -= OnBasisBootModeChanged;
+#endif
             DisableAudioLink();
         }
+
+#if BASIS_FRAMEWORK_EXISTS
+        private void OnBasisBootModeChanged(string mode)
+        {
+            if (_audioLinkEnabled) SetAudioLinkGlobalTexture();
+        }
+#endif
 
         public void UpdateSettings()
         {


pema99:
@towneh Hey, a few questions:

Could this code not just exist in basis itself? Whenever the event fires, you could just call SetAudioLinkGlobalTexture() from the basis codebase, no? Don't really understand why this needs to be in AudioLink.cs.

May this affect other use cases? Is there any way to detect the problematic situation without the basis specific event?

towneh:
@pema99 

If AudioLink prefers not to actively subscribe to Basis Framework events that's completely understandable, but then I would ask that AudioLink takes correct ownership of the lifecycle of their globaltextures as per issue https://github.com/llealloo/audiolink/issues/365 

This is probably new territory for AudioLink, to handle render mode switching, but the flexibility has many advantages.

Short-term Basis can make these calls to satisfy our immediate need for the texture working.

